PR #9, opened 2021-11-01
Changed builing orientation to be North clockwise based.

Hilbrand:
Building orientation is being based on North is 0 degree and clockwise.
For the code in the IMAER-Java project this has limited impact because the building itself is kept as polygon.
But for reading older GML files these should be interpreted as X-as counter-clockwise to be backward compatible.


BertScholten (approved):
LGTM

PR #10, opened 2021-11-05
AER3-1040 generic calculation options

BertScholten:
This PR includes:
- generic calculation options in the Calculation metadata
- utility to convert from CalculationSetOptions to a key-value map
- using that utility to add the used calculation options in the case of calculation results (only adding options if they are not default).
- exporting monitorSrm2Year has been moved to this new functionality.

Since it changes the XSD, the version of IMAER has been updated to 4.0.2.


source/imaer-util/src/main/java/nl/overheid/aerius/util/OptionsMetadataUtil.java
```diff
@@ -0,0 +1,132 @@
...
+    // false should be the default value, so only add if value is true, or if defaults should be added anyway.
+    if (value != null) {
+      mapToAddTo.put(key, value.toString());
+    } else if (addDefaults) {
+      mapToAddTo.put(key, "");
+    }
+  }
+
+  private static void addIntValue(final Map<Option, String> mapToAddTo, final Option key, final int value, final boolean addDefaults) {
+    if (addDefaults || value != 0) {
```
Hilbrand, line 120: This seems a bit tricky, because when 0 is an valid value it's not added (Give the only use case here it seems not a problem now, but might be in the future. Maybe add an option to specify a `no_value` and not add if it matches that value.

Hilbrand (approved):
LGTM

PR #12, opened 2021-11-07
Add new situation types described in AER3-1081

JornC:


BertScholten (approved):
LGTM, though an IMAER change will be required as well at some point

Hilbrand (approved):
LGTM

PR #13, opened 2021-11-08
Remove some obsolete classes and methods.

Hilbrand:


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/EmissionValues.java
```diff
@@ -1,142 +0,0 @@
...
-import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Map.Entry;
-import java.util.Set;
-
-/**
- * Data object to store emission by year/substance.
- */
-public class EmissionValues implements HasEmission, Serializable {
```
BertScholten, line 28: Guess the `HasEmission` method can be removed as well?

BertScholten (approved):

PR #15, opened 2021-11-08
AER3-1127 Correctly grouping standard vehicles for roads

BertScholten:
As it was, the mergingStandardVehicles list would only be kept for one property, which was not terribly useful...
Also ensured that the export order of standard vehicles is more predictable. Especially handy for roundtrip tests.


Hilbrand (approved):
LGTM

PR #16, opened 2021-11-09
Refactored RblCohesionValidator to work with a Scenario instead of Import parcel

Hilbrand:


source/imaer-util/src/main/java/nl/overheid/aerius/validation/RblCohesionValidator.java
```diff
@@ -62,12 +63,20 @@
...
-      addDispersionLineIds(importResult);
-      this.corrections.addAll(importResult.getSituation().getNslCorrections());
+    void add(final ScenarioSituation situation) {
+      addSources(situation);
+      addMeasures(situation);
+      addDispersionLineIds(situation);
+      this.corrections.addAll(situation.getNslCorrections());
+    }
+
+    public void addPoints(final List<CalculationPointFeature> calculationPoints) {
```
BertScholten, line 73: Don't really see a point in making this public since the class is private anyway? Can't hurt though I guess.

BertScholten (approved):

source/imaer-util/src/main/java/nl/overheid/aerius/validation/RblCohesionValidator.java
Hilbrand: As you mentioned the whole class is private so has no impact. To keep all methods the same I've added the `public` to the `add` method.  Other option would be to remove `public` from all other methods.

BertScholten (approved):

PR #19, opened 2021-11-09
AER3-1161 Converting plan emission sources to generic sources on import

BertScholten:
Plan emissionsources are not directly supported anymore. Hence the need to convert these types of sources to something else: generic emission sources.

With this change each activity ends up in it's own separate source. While it might be possible to merge activities into one source if the characteristics are the same, there was probably a reason why the activities were separate in the first place. It's easier for a user to merge the sources if he requires than to split the sources if already merged (because that would require knowing emission factors and original amounts being input).


Hilbrand (approved):
LGTM

PR #20, opened 2021-11-11
Valid gml references

BertScholten:
As it was, the ID's used for references weren't always the same as the ID of the object being referenced, since those objects would receive an ID that was guaranteed to be valid according to the definition (for instance, not starting with a number).

By using the same utility function for those ID's, we should arrive at the same result.


JornC (approved):
ACK!

PR #22, opened 2021-11-23
AER3-1224 Shipping mooring backwards comp. route names

BertScholten:
As it was the label for sources that were created for routes for mooring shipping for older IMAER versions (<= 3.1) were always empty. 

With this change, an attempt is made to get a label for these sources based on their context.
- Inland mooring routes get `[mooring source label]; Route [route index within the mooring source]`
- Maritime Inland mooring routes ('binnengaats') get `[mooring source label]; Route [route index within the mooring source]`
- Maritime maritime mooring routes get `Route [global route index]` (as these routes were re-usable in AERIUS-II between sources, this made a bit more sense).

### DIFF
--- a/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/ship/v31/GML2InlandMooring.java
+++ b/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/ship/v31/GML2InlandMooring.java
@@ -16,6 +16,10 @@
  */
 package nl.overheid.aerius.gml.base.source.ship.v31;
 
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import nl.overheid.aerius.gml.base.AbstractGML2Specific;
 import nl.overheid.aerius.gml.base.GMLConversionData;
 import nl.overheid.aerius.gml.base.IsGmlProperty;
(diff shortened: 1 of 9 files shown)


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/ship/v31/GML2InlandMooring.java
```diff
@@ -95,6 +101,12 @@ private void addRoute(final IsGmlInlandShippingRoute inlandShippingRoute, final
     route.getSubSources().add(imr);
     route.setMooringAId(sourceId);
     addRouteShipsToMooringShips(inlandShippingRoute, vesselGroupEmissionValues);
+    additionalRouteStuff(inlandShippingRoute, route, source);
+  }
+
+  protected void additionalRouteStuff(final IsGmlInlandShippingRoute inlandShippingRoute, final InlandShippingEmissionSource route, final T source)
```
JornC, line 107: Good name!

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/ship/v31/GML2InlandMooring.java
```diff
@@ -131,12 +143,24 @@ private int determinePercentageLaden(final int shipsExisting, final int percenta
...
   }
 
+  protected String constructLabel(final String sourceLabel, final AtomicInteger routeIndexTracker) {
+    // Routes did not have a name in old GML, so come up with our own label
+    // Might not be completely i18n, but works for both NL and EN at least.
+    final int currentIndex = routeIndexTracker.getAndIncrement();
+    final String routeDescription = "Route " + currentIndex;
+    return Stream.of(sourceLabel, routeDescription)
+        .filter(x -> x != null && !x.isBlank())
+        .collect(Collectors.joining("; "));
```
Hilbrand, line 163: Looks like a somewhat complex way of writing. But is a personal preference:
```suggestion
    return (sourceLabel == null || sourceLabel.isBlank() ? "" : (sourceLabel + "; ")) + routeDescription;
```

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/ship/v31/GML2InlandMooring.java
```diff
@@ -95,6 +101,12 @@ private void addRoute(final IsGmlInlandShippingRoute inlandShippingRoute, final
     route.getSubSources().add(imr);
     route.setMooringAId(sourceId);
     addRouteShipsToMooringShips(inlandShippingRoute, vesselGroupEmissionValues);
+    additionalRouteStuff(inlandShippingRoute, route, source);
+  }
+
+  protected void additionalRouteStuff(final IsGmlInlandShippingRoute inlandShippingRoute, final InlandShippingEmissionSource route, final T source)
+      throws AeriusException {
+    // NO-OP for this version
```
Hilbrand, line 109: Can you write some additional comment on why this is a no-op. And with version I assume you refer to the IMAER version.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/ship/v31/GML2InlandMooring.java
BertScholten: yeah, used that construction because I've used it in other places as well. Instead of changing it to your suggestion, moved it to base class so it can be re-used.

Hilbrand (approved):
LGTM

PR #24, opened 2021-11-29
AER3-1238 Mobile source backwards compatibility change

BertScholten:
Where we used to determine operating hours based on fuel consumption for the old category (which also had a possible idle component), we are now doing it based on the fuel consumption of the new category. The assumption is made that the literFuelPerYear is correct (as we don't know if the user specified this based on experience or based on a 'handreiking'). So all we're doing now is converting the liters fuel used to a total number of operating hours, and any idle properties possibly specified by the user are not taken into account.
Numbers used in the unittests are the ones supplied by the 'actualisatieteam' for the corresponding categories.


Hilbrand (approved):
LGTM

PR #27, opened 2021-11-29
Remove deprecated constructor

BertScholten:
Constructor was used to avoid compilation issues in other repository. Should be save to remove after some time (a few days).


Hilbrand (approved):
LGTM

PR #28, opened 2021-11-29
AER3-887 EPSG code for WGS84

BertScholten:
Intended for use with OPS .brn file, where sources can have different SRIDs


Hilbrand:
We can specify the `crs` on a `FeatureCollection`. That might be sufficient for this use case.

BertScholten:
> We can specify the `crs` on a `FeatureCollection`. That might be sufficient for this use case.

Myeah, that's what I went with first. But besides the bigger code change required it also means that you can't mix sources with different SRIDs either (which is possible with the current approach, and I think possible in old AERIUS-II?).

Hilbrand (approved):
LGTM

PR #29, opened 2021-12-13
Refactored out Geometry and Fixed get zoomlevel from id

Hilbrand:
- Removed base class geometry to be able to refactor out the duplicate ReceptorUtil in upstream repository.
- Fixed getZoomLevelForReceptor in ReceptorUtil. Turn out it didn't correctly calculated the zoom level. The method was only used to sort the order in which receptors are calculated this caused it to not be seen earlier.

This is a breaking change.


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/geometry/ReceptorUtil.java
```diff
@@ -236,16 +236,28 @@ public void attachReceptorToGrid(final CalculationPointFeature point) {
...
+    final int rowNumber = (receptorId - 1) / hexHor;
+    final int numberInRow = receptorId - rowNumber * hexHor;
+
+    for (int zoomLevel = 1; zoomLevel < zoomLevelXY.length; zoomLevel++) {
+      final int nextZoomLevel = zoomLevel + 1;
+      final int zoomlevelFactor = (int) Math.pow(2, nextZoomLevel);
+      final int zoomlevelFactorMinusOne = (int) Math.pow(2, zoomLevel);
+
+      if (rowNumber % zoomlevelFactor == 0) {
+        // One of final the rows where final the numbering starts final at the beginning
```
BertScholten, line 249: I don't get what this comment is saying. Should 'final' be removed here?

source/imaer-shared/src/main/java/nl/overheid/aerius/geo/shared/Point.java
```diff
@@ -18,23 +18,19 @@
 
 import java.io.Serializable;
 
-import nl.overheid.aerius.shared.MathUtil;
-
 /**
  * Simple point with x and y coordinate.
  */
-public class Point extends Geometry implements Serializable {
+public class Point extends nl.overheid.aerius.shared.domain.v2.geojson.Point implements Serializable {
```
BertScholten, line 24: Perhaps rename this class to something else, like `SridPoint`? I'd guess we want to prevent to use this class as much as possible if we can, or at least make a clear distinction with the `v2.geojson.Point` version.

BertScholten (changes requested):

source/imaer-shared/src/main/java/nl/overheid/aerius/geo/shared/Point.java
BertScholten: Or perhaps move the class to `AERIUS-II` repository? The `ShippingNode` class is the only one extending the class in `IMAER-java` as far as I can tell, and I don't think we're using that one anymore?

Hilbrand: Yes good idea. I've move the class and renamed it there.

BertScholten (approved):
LGTM

PR #31, opened 2021-12-16
Added option to configure if WNB 25km calculation limit should be applied

Hilbrand:
This to make it possible to either start calculations with or without this limit.

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationSetOptions.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationSetOptions.java
@@ -43,7 +43,7 @@ public class CalculationSetOptions implements Serializable {
   /**
    * If true apply the WNB 25km calculation limit. Default true for WNB.
    */
-  private boolean wnbMaxDistance;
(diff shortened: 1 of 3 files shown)


BertScholten (commented):
Should the option be added to `OptionsMetadataUtil` as well, so it's visible in the GML?

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationSetOptions.java
```diff
@@ -40,6 +40,10 @@
   private boolean forceAggregation;
   private CalculationRoadOPS roadOPS = CalculationRoadOPS.DEFAULT;
   private int monitorSrm2Year;
+  /**
+   * If true apply the WNB 25km calculation limit. Default true for WNB.
+   */
+  private boolean wnbMaxDistance;
```
BertScholten, line 46: Perhaps in this case `useWnbMaxDistance` makes more sense? With just `wnbMaxDistance` I would have expected an integer due to the `Distance` part.

BertScholten (approved):
One suggestion, but LGTM

BertScholten (approved):

PR #32, opened 2022-01-06
Fix adding support for NSLCalculationPoints Type

Hilbrand:
NSLCaclulationPoint can also be part of GML therefor it should be added as type to CalculationPoint.

Fixes AER3-1332 where csv files imported cause an error when starting calculation because csv file is read as NSLCalculationPoint and when receiving at the server as part of a scenario to calculate it can't parse this type object causing it to crash.


BertScholten (approved):
LGTM

PR #33, opened 2022-01-12
Release 4.1.0

BertScholten:


SerhatG (approved):

Hilbrand:
Shouldn't this have been `4.0.2`?

BertScholten:
That is debatable. We were already using `4.1-SNAPSHOT`, so it made some more sense to have it go up. The XSD version isn't necessarily tied to the `IMAER-JAVA` version either (we can change something in the java code that doesn't affect the XSD for instance), so did not see a need to keep that in sync.

There's quite a bit of difference between the released 4.0 and this 4.1.0 version as well, which makes them less compatible. That's why I went with 4.2.0-SNAPSHOT as well as the next development version: if we ever need to update something in the production version, I think we want the option to release a 4.1.1 version that does not contain any changes that have been introduced for newer versions.

However, something can be said to have the first part of the versions in sync. So if the XSD version is 4.0, then the IMAER-java version should be 4.0 as well. Perhaps something to discuss/write down some time?

PR #38, opened 2022-01-13
AER-15 Added Chemistry to OPSOptions

Gijsvanhorn:


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/chemistry/Chemistry.java
```diff
@@ -0,0 +1,25 @@
+package nl.overheid.aerius.shared.domain.chemistry;
+
+public enum Chemistry {
+
+  PROGNOSIS("prognosis", 2018, 2030),
+  ACTUAL("", 2014, 2021);
```
Hilbrand, line 6: Leave the years out if this. It changes every year.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/chemistry/Chemistry.java
```diff
@@ -0,0 +1,17 @@
+package nl.overheid.aerius.shared.domain.chemistry;
```
Hilbrand, line 1: license header

Hilbrand: Move this enum to `OPSOptions` class. It should not be a separate package in `domain` as this is a small feature and that doesn't justify a package on that level.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/OPSOptions.java
```diff
@@ -33,6 +35,7 @@
   private String washout;
   private String convRate;
   private Double roughness;
+  private Chemistry chemistry;
```
Hilbrand, line 38: Chemistry should be processed in `OptionsMetadataUtil`

Hilbrand: Update the `toString`.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/chemistry/Chemistry.java
```diff
@@ -0,0 +1,17 @@
...
+
+public enum Chemistry {
```
Hilbrand, line 3: Can you add some javadoc about what Chemistry represents.

Hilbrand (changes requested):

Hilbrand (approved):
LGTM. Build fails here, but locally it builds.

PR #40, opened 2022-01-13
Added SourceType to ADMS, and additonal safeValueOf methods.

Hilbrand:
Also added more substances and AMBIENT to BuoyancyType.

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/Substance.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/Substance.java
@@ -17,6 +17,7 @@
 package nl.overheid.aerius.shared.domain;
 
 import java.io.Serializable;
+import java.math.BigDecimal;
 import java.util.ArrayList;
 
 /**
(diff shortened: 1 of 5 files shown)


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/Substance.java
```diff
@@ -27,10 +28,30 @@
...
 public enum Substance implements Serializable {
 
+  /**
+   * Zwaveldioxide
+   */
+  SO2(1, "SO2"),
+  /**
+   * Koolstofdioxide
+   */
+  CO2(2, "CO2"),
```
BertScholten, line 38: Is the plan to use these for ADMS as well? Did check that they're in our database definition with the same number, but not sure why they were added there in the first place (or where the numbers originate).

BertScholten (approved):
LGTM, just one question about the substances that were added.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/Substance.java
Hilbrand: I added these as they are available in the database. I don't know yet if these are going to be used. SO2 might be and therefor added them to make the list complete. These number originate from OPS I think. I did try to find the origin once, but couldn't trace it.

PR #41, opened 2022-01-14
Switched to newer JTS library (1.18.2)

BertScholten:
In 2016/2017 JTS moved to locationtech. With version 1.15 (nov 2017) the package names and maven groupIds were changed as well. Think now is a good time to switch along with it (gives some more options as well, which might help with the overlapping hexagon detection for instance). Should contain some performance improvements as well (though not too sure how that'll affect us).
This will probably break applications when using classes like GeometryUtil.


Hilbrand (approved):
LGTM. As you mention, this will probably break the rest of the application due to path changes, so we might need to update downstream right away?

PR #42, opened 2022-01-14
Correct bounds for BNGrid

BertScholten:
The old definition (based on epsg.io) didn't cover england properly (bit in the south-east was cut off). The new defitiniton (based on spatialreference.org, and which is identical to our calculation boundary) does seem to cover UK properly.


Hilbrand (approved):
LGTM

PR #43, opened 2022-01-17
Added interface building limits

Hilbrand:
This way code that checks limits can be independently being developed from OPS and ADMS.

This change introduces the ADMSLimits file in this project.

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/building/BuildingLimits.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/building/BuildingLimits.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright the State of the Netherlands
+ * Crown copyright
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see http://www.gnu.org/licenses/.
+ */
+package nl.overheid.aerius.shared.domain.v2.building;
+
+/**
+ * Interface for Building limits.
+ */
+public interface BuildingLimits {
+
+  default boolean isCircularBuildingSupported() {
+    return false;
+  }
+
+  int buildingDigitsPrecision();
+
+  double buildingHeightMinimum();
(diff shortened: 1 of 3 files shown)


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/building/BuildingLimits.java
```diff
@@ -0,0 +1,47 @@
...
+    return false;
+  }
+
+  int buildingDigitsPrecision();
+
+  double buildingHeightMinimum();
+  double buildingHeightMaximum();
+  double buildingHeightDefault();
+
+
```
BertScholten, line 35: Perhaps a whiteline too much?

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/building/BuildingLimits.java
```diff
@@ -0,0 +1,47 @@
...
+  double buildingWidthMinimum();
+  double buildingWidthMaximum();
+  double buildingWidthDefault();
+
+  double buildingLengthMinimum();
+  double buildingLengthMaximum();
+  double buildingLengthDefault();
+
+  double buildingOrientationMinimum();
+  double buildingOrientationMaximum();
```
BertScholten, line 45: Not sure if it makes much sense to have a min/max for orientation? It's something we determine/define ourselves based on the geometry that is supplied, and not something that the user has to enter.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/building/BuildingLimits.java
```diff
@@ -0,0 +1,47 @@
...
+  int buildingDigitsPrecision();
+
+  double buildingHeightMinimum();
+  double buildingHeightMaximum();
+  double buildingHeightDefault();
+
+
+  double buildingWidthMinimum();
+  double buildingWidthMaximum();
+  double buildingWidthDefault();
```
BertScholten, line 38: Not sure how these defaults are used: our application uses geometry, so we don't really need a default? Or is it possible in ADMS to define a building without defining width/length, and these defaults are used when reading an ADMS specific file?

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/building/BuildingLimits.java
Hilbrand: There is a difference between OPS (179) and ADMS (360). But we don't use it anywhere. So I'll remove it.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/building/BuildingLimits.java
Hilbrand: Hmm. Can be removed. I copied them from ADMS specification. But we don't use it anyway, so I'll remove them.

BertScholten (approved):
LGTM

PR #44, opened 2022-01-17
BuildingLimits should be Serializable to read them through GWT-RPC

Hilbrand:


BertScholten (approved):
LGTM (though not entirely sure where these would be used in GWT-RPC)

Hilbrand:
They're not used yet :smile:, but will be part of the theme configuration.

PR #45, opened 2022-01-18
Added L1/L2 variables to ADMSSourceCharacteristics

Hilbrand:
Add known default for L1, and fixed minimum L1 limit.

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/characteristics/adms/ADMSLimits.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/characteristics/adms/ADMSLimits.java
@@ -68,8 +68,10 @@ public final class ADMSLimits implements BuildingLimits {
   public static final long SOURCE_MASS_FLUX_MAXIMUM = 100_000;
   public static final double SOURCE_MASS_FLUX_DEFAULT = 1;
 
-  public static final long SOURCE_L1_MINIMUM = 0;
+  public static final double SOURCE_L1_MINIMUM = 0.001;
   public static final long SOURCE_L1_MAXIMUM = 1_000;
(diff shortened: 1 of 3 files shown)


BertScholten (approved):
Not a fan of the names `l1` and `l2`. If they can mean different things, for instance if the sourcetype is different, wouldn't it be better to have different properties in our domain model? Wouldn't know how to describe these properties in IMAER for instance.

That being said, if these make sense to the UK people, then I guess we might as well use them...

Hilbrand:
Yes makes sense. The L1 and L2 seem to be used as a kind of generic variables in AMDS. But it does make sense to use specific names.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/characteristics/adms/ADMSLimits.java
```diff
@@ -68,8 +68,10 @@
   public static final long SOURCE_MASS_FLUX_MAXIMUM = 100_000;
   public static final double SOURCE_MASS_FLUX_DEFAULT = 1;
 
-  public static final long SOURCE_L1_MINIMUM = 0;
+  public static final double SOURCE_L1_MINIMUM = 0.001;
   public static final long SOURCE_L1_MAXIMUM = 1_000;
+  public static final long SOURCE_L1_DEFAULT = 1;
```
BertScholten, line 73: Do want to name these with `L1`? It's OK by me if the limits always work for that variable, but if it can differ based on the meaning of the variable...

Hilbrand: These limits are from the documentation where they are written as limits for `L1`. So I kept the reference to L1 for that reason.

PR #47, opened 2022-01-27
Make formatting GML output optional

Hilbrand:
All spaces and newlines can add up when generating large files, therefor to make formatting optional the gml can be generated without formatting.
For testing purposes it's useful to generate formatted output as that makes it easier to read unit test results, therefor it's optional and not fixed.


BertScholten (approved):
LGTM

PR #53, opened 2022-02-04
AER-34 road changes for IMAER 5.0

BertScholten:
* Roads now have a code for area and road type. 
* Vehicle types are no longer an enum, but a String (making them codes as well).
* RoadSpeedProfile is gone as far as IMAER 5.0 is concerned.
* IMAER 5.0 is exported with this PR, though it should be possible to export 4.0 still if that's required (as long as the codes match...)
* Versions of artifacts updated to `5.0-SNAPSHOT` to reflect that IMAER 5.0 is exported.


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/RoadEmissionSource.java
```diff
@@ -22,7 +22,11 @@
 
 public abstract class RoadEmissionSource extends EmissionSourceWithSubSources<Vehicles> {
 
-  private static final long serialVersionUID = 1L;
+  private static final long serialVersionUID = 2L;
+
+  private String roadAreaCode;
+
+  private String roadTypeCode;
```
Hilbrand, line 29: Maybe some java doc to specify the intention of these new fields?

Hilbrand (commented):
I would suggest to rename `imaer:SRM2Road` in v50 too as this specifically refers to the SRM2 model and that is not used for the UK. Or should we introduce a UK specific road emission source, given there are several other differences too. wdyt?

BertScholten:
> I would suggest to rename `imaer:SRM2Road` in v50 too as this specifically refers to the SRM2 model and that is not used for the UK. Or should we introduce a UK specific road emission source, given there are several other differences too. wdyt?

We already have a `imaer:SRM1Road` as well, and both extend `imaer:RoadEmissionSource`. Not too sure what the differences are between NL and UK version, but at least the `tunnelFactor` was already something that did not apply to UK? Think it might be better to have UK specific road emission source in that case.

Hilbrand (approved):
LGTM.
Might indeed be an idea to have a UK specific road source.

PR #56, opened 2022-02-17
AER-262 Fix empty value in results gml when using actual chemistry.

Gijsvanhorn:
The input value is written to gml instead of the label used in the control file to prevent empty values when using ACTUAL chemistry.

Rename field in chemistry enum to make the distinction more clear.


Hilbrand (approved):
LGTM

PR #58, opened 2022-02-22
Removed boolean freeway indicator from SRM2RoadEmissionSource

BertScholten:
It did not really make sense how it was (you could have the urban road sector, but with this property on true?). So better to get rid of it while we're at it.


Hilbrand (approved):
LGTM

PR #61, opened 2022-03-01
Using correct sector ID for default sector characteristics

BertScholten:
As it was, it would use the default characteristics for the ID in the GML, not the one that was possibly converted to the correct sector ID.


Hilbrand (approved):
LGTM

PR #62, opened 2022-03-02
Added ADMSRoadEmissionSource domain object

BertScholten:
New RoadEmissionSource specialization for ADMS. Properties are based on what UI has at the moment.
Renamed RoadSideBarrier to SRM2RoadSideBarrier (and -Type) to reflect that these are specific for SRM2.
Because of the rename, and since the EmissionSourceVisitor has been changed, this will probably require an update in projects using this library.


Hilbrand (approved):
LGTM

PR #63, opened 2022-03-07
AER-180 AER-308 - ADMS Road Barrier defaults

JornC:
Placeholder for road barrier defaults

Also moved the validation constraints here


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/characteristics/adms/ADMSLimits.java
```diff
@@ -84,6 +84,24 @@
...
+  public static final int ROAD_COVERAGE_MAX = 100;
+
+  public static final double ROAD_POROSITY_DEFAULT_NONE = 0D;
+  public static final double ROAD_POROSITY_DEFAULT_NOISE_BARRIER = 5D;
+  public static final double ROAD_POROSITY_DEFAULT_BRICK_WALL = 5D;
+  public static final double ROAD_POROSITY_DEFAULT_STREET_CANYON_TERRACED_HOUSES = 20D;
+  public static final double ROAD_POROSITY_DEFAULT_STREET_CANYON_SEMIDETACHED_HOUSES = 40D;
+  public static final double ROAD_POROSITY_DEFAULT_STREET_CANYON_DETACHED_HOUSES = 50D;
+  public static final double ROAD_POROSITY_DEFAULT_TREE_BARRIER_DENSE = 80D;
+  public static final double ROAD_POROSITY_DEFAULT_TREE_BARRIER_OPEN = 95D;
```
BertScholten, line 103: If these defaults are not bound to be changed anytime soon this is OK for me. But perhaps they could be defined in the enum instead so you don't have to do some switch statement or something when using these?

JornC: Check!

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/road/ADMSRoadSideBarrierType.java
```diff
@@ -21,9 +21,29 @@
...
+  NOISE_BARRIER(5D),
+  BRICK_WALL(5D),
+  STREET_CANYON_TERRACED_HOUSES(20D),
+  STREET_CANYON_SEMIDETACHED_HOUSES(40D),
+  STREET_CANYON_DETACHED_HOUSES(50D),
+  TREE_BARRIER_OPEN(80D),
+  TREE_BARRIER_DENSE(90D),
   OTHER;
+
+  private Double defaultPorosity;
```
Hilbrand, line 33: Does it make sense to use a object instead of primitive? You could use `Double.NaN` for `Other`?

BertScholten (approved):
LGTM (not too sure on `Double` vs `double`, if the object version works for the frontend it's OK I guess)

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/road/ADMSRoadSideBarrierType.java
JornC: We could use primitive, the only real effect would be having to write

```java
if (barrierType.getDefaultPorosity() != Double.NaN) {
  setBarrierLeftPorosity(String.valueOf(v));
}
```

instead of 

```java
Optional.ofNullable(barrierType.getDefaultPorosity())
  .ifPresent(v -> setBarrierLeftPorosity(String.valueOf(v)));
```

I think the latter is cleaner personally

Hilbrand: Ah yes. Still feels strange to not use primitive, and have to rely on null checks. Another option could be to add an extra method `hasPorosity()`.

Hilbrand (approved):

JornC:
Tried using a primitive and using Double.isNaN() rather than a nullcheck, but unfortunately it seems doing these kinds of NaN comparisons does not mesh well with GWT for some reason, so gonna merge the Boxed variety.

PR #65, opened 2022-03-10
AER-324 interpolation for ADMS road emission factors

BertScholten:
Added interpolation mechanism for ADMS roads for standard vehicle types.

Doing both speed and gradient (gradient being part of AER-179, but while at it...)

This will break a project using this library, so needs a PR in our main project to be ready before merging.

### DIFF
--- a/source/imaer-gml/src/test/java/nl/overheid/aerius/test/TestValidationAndEmissionHelper.java
+++ b/source/imaer-gml/src/test/java/nl/overheid/aerius/test/TestValidationAndEmissionHelper.java
@@ -32,6 +32,7 @@ import nl.overheid.aerius.shared.domain.Substance;
 import nl.overheid.aerius.shared.domain.ops.DiurnalVariation;
 import nl.overheid.aerius.shared.domain.v2.characteristics.OPSSourceCharacteristics;
 import nl.overheid.aerius.shared.domain.v2.geojson.Geometry;
+import nl.overheid.aerius.shared.domain.v2.source.road.RoadStandardEmissionFactorsKey;
 import nl.overheid.aerius.shared.domain.v2.source.shipping.inland.InlandWaterway;
 import nl.overheid.aerius.shared.domain.v2.source.shipping.inland.WaterwayDirection;
 import nl.overheid.aerius.shared.domain.v2.source.shipping.maritime.ShippingMovementType;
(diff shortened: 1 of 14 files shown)


source/imaer-gml/src/test/java/nl/overheid/aerius/test/TestValidationAndEmissionHelper.java
```diff
@@ -637,29 +639,32 @@ public boolean isValidRoadSpecificVehicleCode(final String specificVehicleCode)
...
+    return roadStandard(emissionFactorsKey)
         .map(c -> c.stagnatedEmissions.toEmissions())
         .orElseThrow();
   }
 
+  @Override
+  public RoadStandardsInterpolationValues getRoadStandardVehicleInterpolationValues(final RoadStandardEmissionFactorsKey emissionFactorsKey) {
+    // For now just return the supplied values (will result in no interpolation).
+    final int speed = emissionFactorsKey.getMaximumSpeed();
+    final int gradient = Optional.ofNullable(emissionFactorsKey.getGradient()).map(Double::intValue).orElse(0);
```
Hilbrand, line 664: This will round the double down if I'm correct. Is that intended or a problem?

source/imaer-shared/src/test/java/nl/overheid/aerius/shared/emissions/ADMSRoadEmissionsCalculatorTest.java
```diff
@@ -0,0 +1,278 @@
...
+    emissionSource.getSubSources().add(createSpecific());
+    emissionSource.getSubSources().add(createStandard(gradient, true, true));
+
+    ensureSubSourceEmissionsEmpty(emissionSource);
+
+    final Map<Substance, Double> results = emissionsCalculator.calculateEmissions(emissionSource, geometry);
+
+    // Check total emissions
+    // 321.5 * (3703.7010 + 115.0685 + 673.9726 ) / (1000 * 1000)
+    assertEquals(1.4444, results.get(Substance.NOX), 1E-3);
```
Hilbrand, line 91: Preferable an assert should have a message text. Is it an idea to put the commented text above the asserts as part of such message?

source/imaer-gml/src/test/java/nl/overheid/aerius/test/TestValidationAndEmissionHelper.java
BertScholten: For testing purposes it shouldn't matter as we don't really have gradient test values or cases (yet). Might change this later on when working on the gradient story.

source/imaer-shared/src/test/java/nl/overheid/aerius/shared/emissions/ADMSRoadEmissionsCalculatorTest.java
BertScholten: Did add messages to these unittests, but not the commented text above: those are more for checking how to arrive at the expected number (can chug em into Excel or something to see if these values match). I don't think they'll help to clarify anything if the tests fails.

Hilbrand (approved):
LGTM

PR #66, opened 2022-03-14
AER-330 Incorporating direction for ADMS road emissions

BertScholten:
Since it matters for gradient which direction you're going, we better take that into account.
Assumption is made that 50% of vehicles will go from A to B and from B to A if it's the direction BOTH.
Removed the measure bit as well. Not sure if this ever comes into play, but would make the code a bit more complex...


Hilbrand (approved):
LGTM

PR #67, opened 2022-03-15
AER-309 ensuring ADMS road total emission is in kg/year

BertScholten:
Emissions in test case get a bit high, but that's because they're about 10^7 times higher than what we got supplied as actual emission factors.


Hilbrand (approved):
LGTM

PR #68, opened 2022-03-17
AER-308 - Use a default porosity for OTHER

JornC:
> "The type ‘other’ should always have the porosity 0%. At the moment it is 0% when you fill it in immediately, but when you choose another type and then change is to 'other', it keeps the porosity of the previous type. This is confusing for the user."


BertScholten (approved):
LGTM

PR #69, opened 2022-03-21
Add baseline situation type

JornC:


BertScholten (approved):
LGTM (will have to be added to IMAER later on then)

PR #70, opened 2022-03-24
No reuse in tests causing objects to be cached

Hilbrand:
This reuse of forks in the surefire plugin causes objects to be reused over tests.
However these objects can change between tests leading to test to fail if they use the cached version.


BertScholten (approved):
LGTM

PR #73, opened 2022-03-28
Generic code cleanup.

Hilbrand:
- Remove deprecated geo util methods. 
- Reduced XlinkSchema to single class (it's the same for all gml versions).
-  Removed convience constructors on old GML versions. These constructors were used for writing GML, which is not supported for older versions and therefor these constructors can be removed.
-


BertScholten (approved):
LGTM

PR #75, opened 2022-04-01
Fix min monin obukhow length min and default

Hilbrand:
Min should be 1 and default 30 according to ADMS manual.


Hilbrand:
Yes when passing 0 ADMS will stop with an error.

PR #76, opened 2022-04-01
AER-361 circular buildings in IMAER

BertScholten:
Didn't add validation on radius being > 0 when the geometry is a point. Should this be a check at this point, or will it be checked later on anyway?


Hilbrand (approved):
LGTM. If we want a check we might want to add the check to the validation package, and trigger in ImaerImporter like the source validations.

PR #77, opened 2022-04-04
Add width to ADMS source characteristics

Hilbrand:
width on road will be replaced by this, and removed in a follow up pr.

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/characteristics/ADMSSourceCharacteristics.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/characteristics/ADMSSourceCharacteristics.java
@@ -123,11 +123,11 @@ public class ADMSSourceCharacteristics extends SourceCharacteristics {
     this.diameter = diameter;
   }
 
-  public final double getWidth() {
+  public double getWidth() {
     return width;
   }
 
-  public final void setWidth(final double width) {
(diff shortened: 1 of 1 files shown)


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/characteristics/ADMSSourceCharacteristics.java
```diff
@@ -121,6 +123,14 @@ public void setDiameter(final double diameter) {
     this.diameter = diameter;
   }
 
+  public final double getWidth() {
```
BertScholten, line 126: final?

BertScholten (changes requested):
Don't mind the change, but also don't see what this has to do with the width on `ADMSRoadEmissionSource`. That object doesn't use `ADMSSourceCharacteristics` does it?

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/characteristics/ADMSSourceCharacteristics.java
Hilbrand: hmm copied from characteristics.... will remove.

Hilbrand:
You're right. width on road can remain: it doesn't use the characteristics class.

BertScholten (approved):
LGTM

PR #79, opened 2022-04-06
AER-361 ADMS source characteristics

BertScholten:
Introduced ADMS source characteristics in IMAER.

In an IMAER GML you now have the choice between 2 types of source characteristics: `imaer:EmissionSourceCharacteristics` (which is the existing OPS version) and `imaer:ADMSSourceCharacteristics`.

Did not rename `imaer:EmissionSourceCharacteristics`to `imaer:OPSSourceCharacteristics` to keep the amount of changes for the NL version to a minimum (only the order of existing properties changed slightly). This is debatable (the amount of files changed would increase quite a bit due to the existing test files, so would propose to do that in another PR if we want it).

The properties in the GML `ADMSSourceCharacteristics` are missing some that are available in the domain object. Unsure if these should be added as well, but due to the nature of the (binary) EAP file, I'd prefer to add those in a followup PR if needed, as I want to continue work on another part of IMAER.


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/characteristics/GML2ADMSSourceCharacteristics.java
```diff
@@ -0,0 +1,136 @@
...
+
+  public GML2ADMSSourceCharacteristics(final GMLConversionData conversionData) {
+    super(conversionData);
+  }
+
+  @Override
+  protected nl.overheid.aerius.shared.domain.v2.characteristics.ADMSSourceCharacteristics fromGMLSpecific(
+      final IsGmlSourceCharacteristics characteristics,
+      final nl.overheid.aerius.shared.domain.v2.characteristics.ADMSSourceCharacteristics sectorCharacteristics) throws AeriusException {
+    final IsGmlADMSSourceCharacteristics gmlOPSCharacteristics = (IsGmlADMSSourceCharacteristics) characteristics;
```
Hilbrand, line 43: Rename variables to use `ADMS` instead of `OPS` in the name? (e.g. `gmlOPSCharacteristics`). Maybe do a string replacement in this file?

Hilbrand (approved):
LGTM, just some style comments.

PR #81, opened 2022-04-07
AER-361 ADMS road source in IMAER

BertScholten:
Also removed road network from IMAER 5.0 (was unused in 4.0 from AERIUS perspective, now seems a good moment to remove it).


Hilbrand (approved):
LGTM

PR #85, opened 2022-04-08
Removed ImportType and FileFormat

Hilbrand:
These classes are not really relevant for this project as it's focussed on GML only.


BertScholten (approved):
LGTM

PR #86, opened 2022-04-08
GitHub Action: Explicitly add PR information to Sonar analysis

SerhatG:
Normally this should be detected automatically by the Sonar Scanner, but 
since we moved to `pull_request_target` this doesn't work.
The Sonar Scanner might need an update or something else is going on.
By providing it we use the setup as we use for the AERIUS private repo 
in combination with Jenkins.


Hilbrand (approved):
LGTM

PR #88, opened 2022-04-11
AER-353 Overlapping property in IMAER 5.0

BertScholten:
Similar to what was added to 4.0, an optional property to inform about the overlapping 'status' of a receptor.
Currently not reading this property: unsure if this is (ever) required. Shouldn't be too hard to add it if necessary (add to IsGmlReceptorPoint and use it in GML2Result.fillReceptorPoint).


source/imaer-gml/src/main/resources/imaer/5.0/IMAER.xsd
```diff
@@ -2361,7 +2361,13 @@ NOTE Deze klasse beschrijft een rekenpunt op het AERIUS-grid.</documentation>
   <complexType name="ReceptorPointType">
     <complexContent>
       <extension base="IMAER:CalculationPointTypeType">
-        <sequence/>
+        <sequence>
+          <element minOccurs="0" name="overlapping" type="boolean">
+            <annotation>
+              <documentation>In het geval van een berekening met een maximum rekenafstand kan er sprake zijn van een randeffect doordat niet voor alle bronnen rekenresultaten berekend worden op een receptor. Heeft een receptorpunt resultaten van alle bronnen, dan is de overlapping waarde true en is er geen sprake van randeffect. Heeft een receptorpunt niet resultaten van alle bronnen, dan is de overlapping waarde false en is er wel sprake van een randeffect.</documentation>
```
Hilbrand, line 2367: What is the value is `overlapping` is not present in the GML? I assume it means `overlapping` is `true` in that case, as 'overlapping' result is a very specific case. Which is a bit counter intuitive, as in general if no value is present I would assume `false`. I would expect that the points with the `edgeEffect` are the exception and therefor would result in a `true` value. Wouldn't it be more logical to have a property `edgeEffect`?

BertScholten: If the property is absent, that means that the effect has not been determined. I don't think you can safely state that `overlapping` is `true` in that case, as it can also mean that the process to determine this property is not done (for UK version?).

I do agree however that `edgeEffect` as a property might make more sense: `overlapping` is a bit weird concept when receptors/hexagons are concerned (are these hexagons overlapping? no, it's the results that are overlapping). And if you are interested in where the effect occurs, a property that is true is clearer. Then again, you should be interested in the locations where the effect does not occur?

We could rename it to `edgeEffect`, but then we probably should do that consistently. So at least in IMAER-java never use `overlapping`. Wdyt?

Hilbrand: > Then again, you should be interested in the locations where the effect does not occur?

True, but where the effect happens is the exception. So from the data point of view, I would say, it should represent the normal case and the exception should be named. So I would opt for calling it `edgeEffect`, and it would be good to make that consistent.

Hilbrand (approved):
LGTM

PR #92, opened 2022-04-22
Added option enableSubReceptors

Hilbrand:


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/WNBCalculationOptions.java
```diff
@@ -85,6 +87,14 @@ public void setUseReceptorHeights(final boolean useReceptorHeights) {
     this.useReceptorHeights = useReceptorHeights;
   }
 
+  public boolean isSubReceptorsEnabled() {
```
BertScholten, line 90: Not the normal getter name I would expect. Doesn't this give issues when serializing to/from json?

Hilbrand: Yes might be better to keep it in sync.

source/imaer-util/src/main/java/nl/overheid/aerius/util/OptionsMetadataUtil.java
```diff
@@ -109,6 +110,7 @@ private static void wnbOptionsToMap(final Map<Option, String> mapToAddTo, final
     addBooleanValue(mapToAddTo, Option.FORCED_AGGREGATION, options.isForceAggregation(), addDefaults);
     addBooleanValue(mapToAddTo, Option.USE_RECEPTOR_HEIGHT, options.isUseReceptorHeights(), addDefaults);
     addBooleanValue(mapToAddTo, Option.WITH_WNB_MAX_DISTANCE, options.isUseWNBMaxDistance(), addDefaults);
+    addBooleanValue(mapToAddTo, Option.ENABLE_SUB_RECEPTORS, options.isEnableSubReceptors(), addDefaults);
```
BertScholten, line 113: Almost hit the approve, but thought of one more question...

Shouldn't it be `DISABLE_SUB_RECEPTORS` because that would/should be the default behaviour?

Hilbrand: Thought about that too. Saw that stacking also uses inverted option in meta data. So it does make more sense to use inverted. Renamed all to disable.

BertScholten (approved):
LGTM

PR #93, opened 2022-04-26
AER-418 ADMS diurnal variation IMAER, pt 1

BertScholten:
Moved custom diurnal variation field to `SourceCharacteristics` so ADMS version can use it as well.
Introduced new type for custom diurnal variation for the 3-day profile.
Changed the average value as we expect it to 10000.
Changed the domain object to use doubles instead of integers (as discussed).

TODO in next PR: change IMAER to use doubles and incorporate custom diurnal variation in ADMS source characteristics and road objects.


Hilbrand (approved):
LGTM

PR #96, opened 2022-05-03
AER-418 ADMS diurnal variation IMAER, pt 3

BertScholten:
Missing domain part for ADMS road.
Includes a roundtrip test for ADMS road with custom diurnal variation.

Also attempting to avoid some code duplication when it comes to getting/setting these properties.


Hilbrand (approved):
LGTM

PR #97, opened 2022-05-13
Diurnal variation adjustments

BertScholten:
The definitions (and thus the custom diurnal variation profiles) weren't properly exported yet when using `GMLScenarioHelper`.

Also ensured that the main domain ADMS road object uses the properties for diurnal variation available in `ADMSSourceCharacteristics`.


JornC (approved):
ACK

PR #98, opened 2022-05-16
Increase diurnal profile sum tolerance

JornC:


BertScholten (approved):
Still think it's a bit weird to allow this: as far as I understood it the values are used as factors over the emission, so you'd end up adding or removing some emission?

But guess we can always restrict it more later on if required.

PR #99, opened 2022-05-16
AER-510 dynamic srm data

Gijsvanhorn:


Hilbrand (commented):
The duplicated code check won't like this change :smile: Maybe create a SRMVersionOptions class to put the versions in.

Hilbrand (approved):
LGTM

PR #100, opened 2022-05-17
Ensuring definitions get created for ScenarioSituation

BertScholten:
This should prevent the need for nullpointer checks (though the object could still be set to null explicitly somewhere, or be absent when deserializing json).


wtenbosch (approved):
LGTM

PR #101, opened 2022-05-18
Increase BNG zoom level to the max possible

JornC:
Increase max zoom level from 13 to 15, which seems to be the maximum possible (at 16, tiles won't load)


Hilbrand (approved):
LGTM

PR #109, opened 2022-06-23
AER-619 - Add MEDIUM_COMBUSTION_PLANT emission source type

JornC:
AER-619 implements a preview component for "Medium Combustion Plants", this emission source type, which may be changed in some way later, adds a source type for it.


BertScholten (approved):
OK I guess, but doesn't it need a class extending `EmissionSource` as well so the class can be deserialized at the backend?

PR #110, opened 2022-06-24
Add met site terrain properties to ADMSOptions.

Gijsvanhorn:


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/ADMSOptions.java
```diff
@@ -28,6 +28,11 @@ public class ADMSOptions implements Serializable {
   private double minMoninObukhovLength;
   private double surfaceAlbedo;
   private double priestleyTaylorParameter;
+  private int msId;
+  private double msRoughness;
+  private double msMinMoninObukhovLength;
+  private double msSurfaceAlbedo;
+  private double msPriestleyTaylorParameter;
```
BertScholten, line 35: Not sure how this works, but would expect only the `msId` option. And would expect it to be `metSiteId` (or `meteorologicalSiteId` but, guess `met` is a well known abbreviation). Or is it `measurementSiteId`?

I would expect the other options are not something that can be changed by the user, but are based on the site ID. By by adding these to the ADMSOptions object it kind of suggests that there are ways to change these values.

Gijsvanhorn: I've renamed it to `metSiteId`, that's more descriptive and well known enough I think. The other options are here as well, because I need some way to get these values in the adms worker which has no database connection.

BertScholten (approved):
LGTM

PR #112, opened 2022-06-29
AER-197 add ADMS settings to importparcel to enable importing settings.

Gijsvanhorn:


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/importer/ImportParcel.java
```diff
@@ -41,6 +42,7 @@ public class ImportParcel implements Serializable {
   private ScenarioMetaData importedMetaData;
   private final ScenarioSituation situation = new ScenarioSituation();
   private ScenarioSituationResults situationResults;
+  private ADMSOptions admsOptions;
```
Hilbrand, line 45: I would suggest to use `CalculationSetOptions` instead as all other possible options also stored in the gml are from the `CalculationSetOptions` class. To make it more future prove it's better to use the main class instead of the `AMDSOptions` class.

PR #113, opened 2022-06-30
AER-710 Adjusted typo in ADMSRoadSideBarrier in IMAER

BertScholten:
So far only used for UK, which isn't officially released, so should be OK to alter it without changing version (even though technically you could argue we should)


SerhatG (approved):
It won't be an argument from my side. :) LGTM.

PR #114, opened 2022-06-30
AER-690 Properly handling subpoints on import

BertScholten:
As it was, an error would occur when reading a GML containing SubPoints.
Updated GMLReaderTest so it should pick up on errors in the reader during tests from now on.


Hilbrand (approved):
LGTM

PR #115, opened 2022-06-30
AER-709 Update BNGrid bounds

wtenbosch:
Apparently the BNGrid bounds specified here do not contain the entire of Northern Ireland. Updated the bounds to match the calculation boundary. The bounds are only used in OpenLayers. I also removed some settings that we no longer used in AERIUS (and not sure if those are still the correct values...)


BertScholten (approved):
LGTM

PR #116, opened 2022-07-01
Add option to get return default sector when importing GML when sector code unknown

Hilbrand:
Default implementation of isValidSector is true to be compatible with current usages of this library.


BertScholten (approved):
LGTM (though bit hard to tell what the `OPSSourceCharacteristics` change has for effect)

Hilbrand:
The `OPSSourceCharacteristics` should have no effect, as the method interface is compatible. I changed it because the `copyTo` in `OPSSourceCharacteristics` was no override therefor calling this method on an variable defined as the parent class `SourceCharacteristics` would not call the `copyTo` in `OPSSourceCharacteristics` even if the object instance was of type `OPSSourceCharacteristics`. That should not be the case. This change is not directly related to the other changes here. But was done while working or cleaning up Sector, but because that change is a breaking change I didn't do the Sector class changes right now. But I left the fix for the `copyTo` in here.

PR #117, opened 2022-07-04
AER-722 import calc settings from gml

Gijsvanhorn:


source/imaer-util/src/main/java/nl/overheid/aerius/util/OptionsMetadataUtil.java
```diff
@@ -91,6 +94,13 @@ private OptionsMetadataUtil() {
     //util class
   }
 
+  public static void addOptionsFromMap(final Map<Option, String> map, final CalculationSetOptions options) {
+    options.setStacking(!Boolean.parseBoolean(map.get(Option.WITHOUT_SOURCE_STACKING)));
+
+    final NCACalculationOptions ncaCalculationOptions = options.getNcaCalculationOptions();
```
Hilbrand, line 100: The NCA options should be conditional based on the theme, like in `optionsToMap`.

Hilbrand (changes requested):

Hilbrand (approved):
LGTM

PR #123, opened 2022-07-11
AER-722 Made GMLCalculationSetOptionsReader more robust

BertScholten:
It should fail less on parts of the structure that are option (for instance, max range is optional, and shouldn't result in an error when not available).
Also included some tests that show behaviour (some of which were failing with the old code).
Also removed reading the stacking option, as I don't think it's something that we want to import from GML (for now at least).


Gijsvanhorn (approved):
LGMT, just a small remark

PR #124, opened 2022-07-22
AER-795 serialize the list of NSL corrections so they can be included in the calculation.

Gijsvanhorn:
The corrections are needed when serializing because the scenario is stored as a file when sending the calculation task.


BertScholten (approved):
LGTM (this didn't change any of the unittests? would have expected a change in the json roundtrip test files)

Gijsvanhorn:
> LGTM (this didn't change any of the unittests? would have expected a change in the json roundtrip test files)

Apparently for the roundtrip tests the objectmapper doesn't serialize empty values. As NSLMeasures is an empty array it doesn't get serialized and the tests didn't fail. 

I'm going to add some test content for the roundtrip test.

PR #126, opened 2022-08-03
Enable serialization of nsl measures to json

Gijsvanhorn:
because we need those to transfer scenario data from connect to the worker via the fileservice.


JornC (approved):
ACK

PR #127, opened 2022-08-04
AER-417 Out of range year fix

MichielJanssen-DAT:
- Don't update year from metadata if set already
- Update warning for changing year

NOTE: I'm not sure if this is the best way, but this is the best way I found
Linked PR: https://github.com/aerius/AERIUS-II/pull/4322

### DIFF
--- a/source/imaer-gml/src/main/java/nl/overheid/aerius/importer/ImaerImporter.java
+++ b/source/imaer-gml/src/main/java/nl/overheid/aerius/importer/ImaerImporter.java
@@ -283,9 +283,12 @@ public class ImaerImporter {
   private void setImportResultMetaData(final ImportParcel result, final GMLReader reader) {
     final GMLMetaDataReader metaDataReader = reader.metaDataReader();
     result.setImportedMetaData(metaDataReader.readMetaData());
-    result.getSituation().setYear(metaDataReader.readYear());
     result.setVersion(metaDataReader.readAeriusVersion());
     result.setDatabaseVersion(metaDataReader.readDatabaseVersion());
+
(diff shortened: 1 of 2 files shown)


MichielJanssen-DAT:
Die gefaalde run komt vast door die WIP exceptie

source/imaer-gml/src/main/java/nl/overheid/aerius/importer/ImaerImporter.java
```diff
@@ -283,9 +283,14 @@ private static ArrayList<CalculationPointFeature> filterResultPoints(final List<
...
     result.setImportedMetaData(metaDataReader.readMetaData());
-    result.getSituation().setYear(metaDataReader.readYear());
     result.setVersion(metaDataReader.readAeriusVersion());
     result.setDatabaseVersion(metaDataReader.readDatabaseVersion());
+
+    if (result.getSituation().getYear() == 0) {
+      result.getSituation().setYear(metaDataReader.readYear());
+    } else {
+      result.getWarnings().add(new AeriusException(ImaerExceptionReason.GML_METADATA_YEAR_NOT_WITHIN_RANGE, metaDataReader.readYear()));
+    }
```
BertScholten, line 293: This seems incorrect: if the year in the importparcel wasn't 0, it'll always trigger a warning, even if the year to be imported was actually within the range (or even the same as the year already present in the situation).

Don't think this should be handled in this project anyway, but solely in the AERIUS-II project: this library shouldn't care about ranges and such when importing, it should just return what is read from the file.

BertScholten (changes requested):

PR #128, opened 2022-08-09
AER-760 user scenario added to GML roundtriptest

johnverberne:
add 7 user scenario as GML to GML roundtriptest


BertScholten (approved):
LGTM

PR #129, opened 2022-08-12
AER-651 Added a warning when netting factor not supplied for netting situation

BertScholten:
Added a test case in the GMLValidateErrorsTest for this one as well (might have expected it in GMLValidateWarningsTest, but tbh that one should be refactored a bit).


source/imaer-gml/src/test/java/nl/overheid/aerius/gml/GMLValidateErrorsTest.java
```diff
@@ -201,6 +201,12 @@ public void testGMLRoadSegmentNotAFraction() throws IOException {
         e -> assertEquals(50D, Double.parseDouble(e.getArgs()[0]), EPSILON, "Expected road segment position to be 50"));
   }
 
+
```
MichielJanssen-DAT, line 204:
```suggestion
```

MichielJanssen-DAT (approved):
LGTM

PR #130, opened 2022-08-12
Got rid of old offroad vehicle specification code

BertScholten:
It's still around in 5.0 XSD, but we aren't using it and it's up for removal.


JornC (approved):
ACK

Hilbrand (approved):
LGTM

PR #131, opened 2022-08-22
Fix issue with creating polygons with holes

BertScholten:
As it was, an ArrayIndexOutOfBoundsException would occur when a polygon with any number of holes would be supplied when converting from AERIUS to JTS objects.

### DIFF
--- a/source/imaer-util/src/test/java/nl/overheid/aerius/util/GeometryUtilTest.java
+++ b/source/imaer-util/src/test/java/nl/overheid/aerius/util/GeometryUtilTest.java
@@ -179,7 +179,6 @@ class GeometryUtilTest {
 
       @Override
       public GeometryType type() {
-        // TODO Auto-generated method stub
         return null;
       }
     };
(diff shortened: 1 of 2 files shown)


source/imaer-util/src/test/java/nl/overheid/aerius/util/GeometryUtilTest.java
```diff
@@ -145,6 +147,62 @@ void testToConvexHull() throws AeriusException {
...
+    final Geometry convertedPolygonWithHole = GeometryUtil.getGeometry(polygonWithHole);
+
+    assertTrue(convertedPolygonWithHole instanceof org.locationtech.jts.geom.Polygon, "Polygon with hole should convert to JTS polygon");
+
+    final nl.overheid.aerius.shared.domain.v2.geojson.Geometry anonymousGeometry = new nl.overheid.aerius.shared.domain.v2.geojson.Geometry() {
+      private static final long serialVersionUID = 1L;
+
+      @Override
+      public GeometryType type() {
+        // TODO Auto-generated method stub
```
JornC, line 182:
```suggestion
```

JornC (approved):
ACK but see comment

Hilbrand (approved):
LGTM

PR #132, opened 2022-08-25
AER-685 Show warning when building height is zero or less

johnverberne:


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/exception/ImaerExceptionReason.java
```diff
@@ -441,6 +441,13 @@ public enum ImaerExceptionReason implements Reason {
    */
   COHESION_REFERENCE_MISSING_BUILDING(5521),
 
+  /**
+   * Buildingheight is < 0.
+   *
+   * @param 0 ID of the building that has height < 0.
```
BertScholten, line 447:
```suggestion
   * Buildingheight is <= 0.
   *
   * @param 0 Label of the building that has height <= 0.
```

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/exception/ImaerExceptionReason.java
```diff
@@ -441,6 +441,13 @@ public enum ImaerExceptionReason implements Reason {
...
+   */
+  BUILDING_HEIGHT_ZERO(5522),
```
BertScholten, line 449: Not sure about the name, perhaps `BUILDING_HEIGHT_TOO_LOW`?

BertScholten, line 449: Most of the 55XY numbers are currently about cohesion. Might make more sense to add this one to the 52XY range?

BertScholten (changes requested):
Like the addition of the unittest. Could also add a test case based on a GML file that would cause the error to `GMLValidateErrorsTest`, but that's optional with the adition of the util unittest as far as I'm concerned.

BertScholten (approved):
LGTM

PR #133, opened 2022-08-26
Generate a sane id by removing the 0. part of the generated number.

Hilbrand:
Getting a string id with "0." feels weird therefor remove that part.


BertScholten (approved):
OK... I guess?

Don't really see a problem with the `0.` bit tbh, as long as it's unique who cares that it starts with that? And the benefit of the `0.` bit is that people won't be tempted to use it as an actual user-facing identifier somewhere. And the front-end is using some `DOM.createUniqueId` string at the moment, which is OK as well, but if someone would change this to a number field...

Would rather get rid of the `Math.random`, but think there were some issues with using UUID in combination with GWT iirc.

PR #135, opened 2022-08-30
Versions checking and updating

BertScholten:
Version checking using `versions-maven-plugin` and `dependency-check-maven`.

Also made some plugin versions explicit.

Also updated some versions based on feedback from these.

WIP till after next release.


Hilbrand (approved):
LGTM

Hilbrand (approved):
LGTM

PR #136, opened 2022-08-30
AER-790 Move generic Geo classes to a separate module

Gijsvanhorn:


Hilbrand:
Maybe keep all the package structure for all files, so we don't have to update the depending projects.

Gijsvanhorn:
> Maybe keep all the package structure for all files, so we don't have to update the depending projects.

sure, that makes this a lot simpler.

Hilbrand (approved):
LGTM

PR #140, opened 2022-09-09
Simplified EPSG by introducing an new enum variant

Hilbrand:
EPSG only needed to contain the SRID code. The other properties are specific for the UI and not needed here. Kept backward compatible for now. When this change is added the depending code can start using the new enum

### DIFF
--- a/source/shared-geo/src/main/java/nl/overheid/aerius/shared/geo/EPSG.java
+++ b/source/shared-geo/src/main/java/nl/overheid/aerius/shared/geo/EPSG.java
@@ -41,7 +41,7 @@ public enum EPSG {
    */
   TM75(29903);
 
-  private static final String EPSG_PRE_TEXT = "EPSG:";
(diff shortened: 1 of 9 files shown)


source/shared-geo/src/main/java/nl/overheid/aerius/shared/geo/EPSG.java
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Irish grid (EPSG:29902).
+   */
+  TM65(29902),
+  /**
+   * Northern Irish grid (EPSG:29903).
+   */
+  TM75(29903);
+
+  private static final String EPSG_PRE_TEXT = "EPSG:";
```
Gijsvanhorn, line 44:
```suggestion
  private static final String EPSG_PREFIX = "EPSG:";
```

Gijsvanhorn (approved):
LGTM, just a small rename suggestion

PR #144, opened 2022-09-19
AER-858 don't abuse the farmland code for the type. 

Gijsvanhorn:


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/FarmlandEmissionsCalculator.java
```diff
@@ -35,38 +37,37 @@ public FarmlandEmissionsCalculator(final FarmlandEmissionFactorSupplier farmland
...
+              .multiply(BigDecimal.valueOf(activity.getNumberOfAnimals())), (v1, v2) -> v1.add(v2))
+      );
+    }
+
+    if (activity.getFarmEmissionFactorType() == FarmEmissionFactorType.PER_ANIMAL_PER_YEAR) {
+      emissionFactors.forEach((key, value) ->
+          summedEmissions.merge(key, BigDecimal.valueOf(value).multiply(BigDecimal.valueOf(activity.getNumberOfAnimals())), (v1, v2) -> v1.add(v2))
       );
-    } else {
-      updateEmissions(activityEmissionFactors, summedEmissions);
```
BertScholten, line 69: might be an idea to use `else if` and a final `else` clause where you throw an exception if the type is unknown?

source/imaer-shared/src/test/java/nl/overheid/aerius/shared/emissions/FarmlandEmissionsCalculatorTest.java
```diff
@@ -29,16 +30,15 @@
...
+import nl.overheid.aerius.shared.domain.v2.source.farmland.AbstractFarmlandActivity;
+import nl.overheid.aerius.shared.domain.v2.source.farmland.CustomFarmlandActivity;
+import nl.overheid.aerius.shared.domain.v2.source.farmland.StandardFarmlandActivity;
 import nl.overheid.aerius.shared.exception.AeriusException;
 
 @ExtendWith(MockitoExtension.class)
 class FarmlandEmissionsCalculatorTest {
 
-  private static final String GRAZING_CATEGORY_CODE = "grazingCategoryCode";
+  private static final String PASTURE_CATEGORY_CODE = "pastureCategoryCode";
```
BertScholten, line 41: could use `CATEGORY_CODE = "someCategoryCode"` to avoid the pasture vs grazing discussion here ;)

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/FarmlandEmissionsCalculator.java
```diff
@@ -35,38 +37,37 @@ public FarmlandEmissionsCalculator(final FarmlandEmissionFactorSupplier farmland
...
-    Map<Substance, Double> userEmissions = activity.getEmissions();
-
-    // Prefer emission factors supplied by the farmlandEmissionFactorSupplier
-    Map<Substance, Double> activityEmissionFactors = repositoryEmissions.isEmpty() ? userEmissions : repositoryEmissions;
+  @Override
+  public void visit(final StandardFarmlandActivity activity, final Map<Substance, BigDecimal> summedEmissions) {
+    final Map<Substance, Double> emissionFactors = farmlandEmissionFactorSupplier
+        .getFarmSourceEmissionFactors(activity.getFarmSourceCategoryCode());
 
     if (activity.getFarmEmissionFactorType() == FarmEmissionFactorType.PER_ANIMAL_PER_DAY) {
```
BertScholten, line 60: The `FarmEmissionFactorType` shouldn't be present on the activity I think, but it should be retrieved from the `farmlandEmissionFactorSupplier` based on both the `activityCode` and the `farmSourceCategoryCode`? It's not something that is defined by the user, but rather by the definition in the database for the supplied activity code (which translates to some sector) and the actual category.

BertScholten (changes requested):

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/FarmlandEmissionsCalculator.java
Gijsvanhorn: Good idea, I think the `farmSourceCategoryCode` will be enough, since the code is unique in table `farm_source_categories` and each row must contain the emissionFactorType.

source/imaer-shared/src/test/java/nl/overheid/aerius/shared/emissions/FarmlandEmissionsCalculatorTest.java
Gijsvanhorn: grazing was in more places in this test, even though it's not relevant anymore. Its just custom and standard now.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/FarmlandEmissionsCalculator.java
BertScholten: ah yeah, if that's guaranteed to be unique that should work as well, thought there might be a pkey on sector/code combination, but this works as well.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/FarmlandEmissionsCalculator.java
```diff
@@ -22,51 +22,55 @@
...
+          summedEmissions.merge(key, BigDecimal.valueOf(value).multiply(BigDecimal.valueOf(activity.getNumberOfDays()))
+              .multiply(BigDecimal.valueOf(activity.getNumberOfAnimals())), BigDecimal::add)
+      );
+    } else if (emissionFactorType == FarmEmissionFactorType.PER_ANIMAL_PER_YEAR) {
+      emissionFactors.forEach((key, value) ->
+          summedEmissions.merge(key, BigDecimal.valueOf(value).multiply(BigDecimal.valueOf(activity.getNumberOfAnimals())), BigDecimal::add)
       );
     } else {
-      updateEmissions(activityEmissionFactors, summedEmissions);
+      throw new AeriusException(ImaerExceptionReason.INTERNAL_ERROR, "Invalid farm emission factor type for StandardFarmlandActivity.");
```
BertScholten, line 73: Was intended to be `GML_FARM_EMISSION_FACTOR_TYPE_UNKNOWN` instead of the internal error I guess (else that one was added for nothing). 

Come to think of it, that would be more an indication that the `farmSourceCategoryCode` was unknown in the first place if it reaches a user, so it should perhaps be captured by a validation part (but can do that later on with the actual IMAER changes), and then an internal error might make sense here: It would only reach this in case of a new type that we didn't implement fully yet, or if nothing was validated.

BertScholten (changes requested):

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/FarmlandEmissionsCalculator.java
Gijsvanhorn: Right, looking at other EmissionCalculators throwing an internal error in this case is the expected behavior. I'll remove `GML_FARM_EMISSION_FACTOR_TYPE_UNKNOWN`

BertScholten (approved):
LGTM

PR #145, opened 2022-09-20
Added a undefined 'additionalOptions' to CalculationSetOptions

BertScholten:
This can be used by a program using this library to specify their own options without having to change anything in this library to get that option included in the whole Scenario structure.

For example, when using Jackson, you can use MixIn to specify an exact type to handle deserialization in the correct class, and then you can set and get these additional options (though you'll still have to cast when getting them).


Hilbrand (approved):
LGTM

PR #146, opened 2022-09-22
IMAER 5.1(.0)

BertScholten:
In preparation of farmland related changes, which most likely will be backwards compatible (if not, we can always switch it to 6.0).


Hilbrand (approved):
LGTM

We might think about if we need to support writing to different versions? One for stable release and one for latest. Such that when using stable it still uses the released version, and latest the new version that is still in development. But it might have more consequences for the backend that make supporting 2 versions to complex. Like what to do with newly added features not in the stable. These should not be used when running stable version.

BertScholten:
myeah, I guess we could do something with that, so you get to pick which version to export when using the library (default being the one that matches the library version). You'd have to ensure not to enable new features in stable in the UI then as well.

PR #148, opened 2022-09-28
FarmlandActivity emission calculations and bugfix

BertScholten:
* Emissions weren't set for the Farmland subsources (FarmlandActivity).
* `numberOfDays ` were erased when they were expected and specified. Now only erased when they're not expected.

AER-868, AER-873, AER-876


johnverberne (approved):
LGTM

PR #149, opened 2022-09-29
Changed error codes to be more consistent

BertScholten:
The unknown farm emission factor type is purely something that can occur when importing GML (in our domain code it's already an enum, so harder to get wrong). The other reasons are triggered during normal validation, and can also occur on wrong input within our application (unexpected UI behaviour).

Also turns out the wrong reason was used...


johnverberne (approved):
LGTM

PR #150, opened 2022-10-03
Removed deprecated EPSG classes

Hilbrand:


BertScholten (approved):
LGTM, might be best to wait with merging this for a few days though, to allow other repositories to catch up

BertScholten:
Think it has been a few days, and with recent release now seems a good time to merge this.

PR #151, opened 2022-10-12
AER-962 added road local fraction NO2 to custom calculation points

BertScholten:


Hilbrand (changes requested):
I would prefer a new subclass instead of using customcalcuationpoint. Although I don't know a good name.

BertScholten:
I guess we could do that, but then you'd have to create the correct type in the front end when making a new custom point, or in backend when creating custom points automagically.

I guess `NcaCalculationPoint` would work, or `ADMSCalculationPoint`? Though not sure if this is directly tied to ADMS.

Hilbrand (approved):
LGTM

PR #152, opened 2022-10-12
AER-965 road local fraction NO2 in NCACalculationOptions

BertScholten:
Also included some roundtrips to show how this would work, the GML requiring some more changes as we didn't have a roundtrip for NCA calculation options yet.


Hilbrand (approved):
LGTM

PR #153, opened 2022-10-14
New scail emission factor types

BertScholten:
Also changed validation so these can be validated in a generic approach.
Not too sure about the interface with all the default methods, but it does seem to work based on the unittests...


wtenbosch (approved):
LGTM

PR #154, opened 2022-10-17
Introduced a CalculationPointDuplicator

BertScholten:
Used the CalculationPointVisitor, that has been updated with some missing types, so using `calculationPointFeature.accept(duplicator)` should be enough to duplicate a point (without it's results). Methods are protected with the idea that this should make it easier to override any of the types, in which case this class can be extended for the appropriate type (no longer visitor pattern in that case, but wouldn't know how else to do that with types not available in this library).


Hilbrand (approved):
LGTM

PR #156, opened 2022-10-20
AER-856 manure storage 

BertScholten:
New emission source object type: `ManureStorageEmissionSource`. Uses the whole `FarmEmissionFactorType` approach to do calculations/validations.

Also includes some changes for farmland activities to support some more emission factor types.


source/imaer-util/src/test/java/nl/overheid/aerius/validation/ManureStorageValidatorTest.java
```diff
@@ -0,0 +1,226 @@
...
+
+    final List<AeriusException> errors = new ArrayList<>();
+    final List<AeriusException> warnings = new ArrayList<>();
+    final ManureStorageValidator validator = new ManureStorageValidator(errors, warnings, validationHelper);
+
+    final boolean valid = validator.validate(source);
+
+    assertTrue(valid, "Valid test case");
+    assertTrue(errors.isEmpty(), "No errors");
+    assertTrue(warnings.isEmpty(), "No warnings");
```
Hilbrand, line 64: You could see if you can put the validation/assert into a method with parameters to reduce duplication of code.

Hilbrand (approved):
LGTM. Just a comment about reducing code in unit test.

source/imaer-util/src/test/java/nl/overheid/aerius/validation/ManureStorageValidatorTest.java
BertScholten: Good point, see latest commit.

PR #157, opened 2022-10-21
Split road local fraction NO2 into receptors and points specific options

Hilbrand:


source/imaer-util/src/main/java/nl/overheid/aerius/util/OptionsMetadataUtil.java
```diff
@@ -174,24 +175,26 @@ private static void opsOptionsToMap(final OPSOptions options, final Map<Option,
...
+    if (options.getRoadLocalFractionNO2ReceptorsOption() == RoadLocalFractionNO2Option.ONE_CUSTOM_VALUE
+        || options.getRoadLocalFractionNO2PointsOption() == RoadLocalFractionNO2Option.ONE_CUSTOM_VALUE) {
       options.setRoadLocalFractionNO2(
           Optional.ofNullable(map.get(Option.ROAD_LOCAL_FRACTION_NO2_CUSTOM_VALUE)).map(Double::parseDouble).orElse(null));
     }
     parseMeteoYears(options, map);
 
     options.getAdmsOptions()
-        .setMinMoninObukhovLength(getOrDefault(map, Option.ADMS_MIN_MONIN_OBUKHOV_LENGTH, ADMSLimits.MIN_MONIN_OBUKHOV_LENGTH_DEFAULT));
+    .setMinMoninObukhovLength(getOrDefault(map, Option.ADMS_MIN_MONIN_OBUKHOV_LENGTH, ADMSLimits.MIN_MONIN_OBUKHOV_LENGTH_DEFAULT));
```
BertScholten, line 188: Weird indentation changes in this file

BertScholten (approved):
LGTM, though I thought one option for both would work, and in the case of `INDIVIDUAL_CUSTOM_VALUES` for receptors we'd use `NaN` or something as we'd do for custom points that didn't have a custom value specified. Also, think the option INDIVIDUAL_CUSTOM_VALUES doesn't make sense for receptors anyway?

PR #158, opened 2022-10-21
AER-1065 duplicate coordinate check

BertScholten:
As it turns out, the intersection check implementation was changed with the switch to a newer JTS version. This caused duplicate-points-within-a-polygon to trigger the invalid geometry due to intersections error.

This change reverts that behaviour: duplicate-points-within-a-polygon will no longer be seen as invalid.


Hilbrand (approved):
LGTM

PR #159, opened 2022-10-25
Allow emission factor type PER_YEAR for manure storage

BertScholten:
For custom manure storage it should be possible to specify a kg/y emission factor (in which case it's also directly the total kg/y emission).


Hilbrand (approved):
LGTM

PR #161, opened 2022-11-07
AER-816 Added CUSTOM_RECEPTORS as CalculationType

wtenbosch:
Required by https://github.com/aerius/AERIUS-II/pull/4772

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationType.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationType.java
@@ -29,6 +29,10 @@ public enum CalculationType {
    */
   CUSTOM_POINTS,
   /**
+   * Calculate with custom receptors (maps user defined points to receptors on the predefined grid).
+   */
+  CUSTOM_RECEPTORS,
+  /**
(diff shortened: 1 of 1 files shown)


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationType.java
```diff
@@ -28,6 +28,10 @@ public enum CalculationType {
    * Calculate with custom points (fixed set of receptors or user defined points).
    */
   CUSTOM_POINTS,
+  /**
+   *
```
BertScholten, line 32: Could use a small description as to what the option means (Calculates using custom points mapped to receptors or something like that?)

BertScholten (changes requested):

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationType.java
wtenbosch: Oh wow how did I miss that, I guess I quickly added this and thought that I would add a description later.

BertScholten (approved):
LGTM

PR #162, opened 2022-11-07
AER-941 GML traceability

BertScholten:
Or rather, including information about other situations that were included in a calculation when including results.
While the IMAER definition includes a reference field for 'OtherSituationMetadata', this won't be exported (yet): currently the reference is always generated during GML generation, and the reference is supplied through the ScenarioMetaData, which would be the same for all situations... So if that's a critical requirement, that requires some more work.


source/imaer-gml/src/main/resources/imaer/5.1/IMAER.xsd
```diff
@@ -2361,6 +2366,36 @@ NOTE Voor het bepalen van de emissie van mobiel werktuig kan gekeken worden naar
...
+        <annotation>
+          <documentation>Type van de andere situatie.</documentation>
+        </annotation>
+      </element>
+      <element minOccurs="0" name="name" type="string">
+        <annotation>
+          <documentation>Naam van de andere situatie.</documentation>
+        </annotation>
+      </element>
+      <element minOccurs="0" name="reference" type="string">
```
Hilbrand, line 2387: Shouldn't `reference` not be required? just putting in an other situation with only situationType isn't useful information? (name might also be something to make required?)

BertScholten: This mirrors how it is in SituationMetadata. Seemed appropriate, as technically it's possible to have a GML without a reference or name, and then you couldn't use this object to refer to that GML. It's also something that we export, and not something that we read, so we're in charge of making sure it's filled in anyway. 

Even without reference/name the main goal of this new metadata object, indicating that there are other situations used in a calculation, is still achieved. It's less useful, but not totally useless.

Also, with the current state of this PR `reference` will not be filled in by AERIUS: references are generated during the GML generation proces (so bit of a order problem there), and references are kept on Scenario level, not on Situation level (due to it being generated during GML proces you end up with different references per situation, but in our current model it's defined per Scenario).

Hilbrand (approved):
LGTM

PR #164, opened 2022-11-10
Add the reference to the situation during import

wtenbosch:
It seems like the reference (on the situation) is not set during import. But maybe this should be fixed somewhere else?


BertScholten (commented):
Do we use the imported reference anywhere?

wtenbosch:
> Do we use the imported reference anywhere?

Yes, in the downloadable log with errors and warnings. To reference which file caused the error or warning. That was what was causing the HTTP 500 with downloading the log (there also was a nullpointer guard, but that did not work anymore, I am fixing that as well in AERIUS-II)

BertScholten (approved):
Ah ok, missed that. Guess it can't hurt to import it again for that, just thought it wasn't really used.

PR #165, opened 2022-11-15
AER-1124 Only set nettingFactor if situationType == SituationType.NETTING

wtenbosch:


BertScholten (approved):
LGTM (thought I had approved this one already)

PR #166, opened 2022-11-15
Read missing calculation metadata as null

BertScholten:
Slight change in contract here, that this method can now return null instead of a new object. Benefit of this is that it's now clear when the imported parcel did not have any calculation metadata to begin with.


Hilbrand (approved):
LGTM

PR #170, opened 2022-11-21
Added validations to replace schematron

BertScholten:
These are validations that are currently handled by schematron. If we want to remove schematron, we better make sure those validations are still around.

Also deprecated the classes that are most likely used outside of the library when it comes to schematron use.

This is step 1 when it comes to removing schematron.

### DIFF
--- a/source/imaer-util/src/main/java/nl/overheid/aerius/validation/NSLMeasureValidator.java
+++ b/source/imaer-util/src/main/java/nl/overheid/aerius/validation/NSLMeasureValidator.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright the State of the Netherlands
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


source/imaer-util/src/main/java/nl/overheid/aerius/validation/NSLMeasureValidator.java
```diff
@@ -0,0 +1,56 @@
...
+    for (final NSLMeasureFeature feature : measureFeatures) {
+      final NSLMeasure measure = feature.getProperties();
+      final Optional<Double> negativeValue = measure.getVehicleMeasures().stream()
+          .flatMap(x -> x.getEmissionReductions().stream())
+          .map(EmissionReduction::getFactor)
+          .filter(x -> x < 0)
+          .findFirst();
+      if (negativeValue.isPresent()) {
+        exceptions.add(new AeriusException(ImaerExceptionReason.UNEXPECTED_NEGATIVE_VALUE, measure.getLabel(), String.valueOf(negativeValue.get())));
+      }
```
Hilbrand, line 52:
```suggestion
      measure.getVehicleMeasures().stream()
          .flatMap(x -> x.getEmissionReductions().stream())
          .map(EmissionReduction::getFactor)
          .filter(x -> x < 0)
          .findFirst()
          .ifPresent(negativeValue -> exceptions.add(new AeriusException(ImaerExceptionReason.UNEXPECTED_NEGATIVE_VALUE, measure.getLabel(), String.valueOf(negativeValue));
```

Hilbrand (approved):
LGTM, just a small comment.

Hilbrand (approved):

PR #172, opened 2022-12-14
AER-824 building validation more strict

BertScholten:
Also some other validation related changes, like moving the CustomDiurnalVariation validation and introducing a new Validator that can help with validation a whole situation. 
That last class is intended for last-minute checks, like ensuring the objects we're about to calculate are actually valid (we already do this, but for sources only).


wtenbosch (approved):
LGTM

wtenbosch:
There is a test failure btw: GMLValidateErrorsTest.testGMLCustomDiurnalVariationSum:249->assertResult:269->lambda$testGMLCustomDiurnalVariationSum$15:253 Expected sum ==> expected: <24.0> but was: <24>

PR #178, opened 2023-02-03
Wrap invalid geometries to throw an AeriusException instead.

Hilbrand:
Added unit test. (Removed dependency on WKTGeometry too)

### DIFF
--- a/source/imaer-util/src/main/java/nl/overheid/aerius/util/GeometryUtil.java
+++ b/source/imaer-util/src/main/java/nl/overheid/aerius/util/GeometryUtil.java
@@ -127,7 +127,11 @@ public final class GeometryUtil {
       final Coordinate coordinate = toJtsCoordinate(aeriusLineString.getCoordinates()[i], aeriusLineString);
       coordinates[i] = coordinate;
     }
-    return geometryFactory.createLineString(coordinates);
+    try {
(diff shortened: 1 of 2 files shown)


source/imaer-util/src/main/java/nl/overheid/aerius/util/GeometryUtil.java
```diff
@@ -127,7 +127,11 @@ private static LineString toJtsLineString(final nl.overheid.aerius.shared.domain
       final Coordinate coordinate = toJtsCoordinate(aeriusLineString.getCoordinates()[i], aeriusLineString);
       coordinates[i] = coordinate;
     }
-    return geometryFactory.createLineString(coordinates);
+    try {
+      return geometryFactory.createLineString(coordinates);
+    } catch (final IllegalArgumentException e) {
+      throw new AeriusException(ImaerExceptionReason.GEOMETRY_INVALID, e.getMessage());
```
BertScholten, line 133: The use `e.getMessage()` might be a bit inconsistent: for other locations where this reason is used we use the supplied geometry. So error messages with this might end up looking a bit weird?

BertScholten (changes requested):

BertScholten (approved):
LGTM

PR #180, opened 2023-02-17
AER-1377 Introduced CalculationMethod as replacement for CalculationType

Hilbrand:
CalculationMethod will be the new name for CalculationType. Older GML files will still have the calculation type, but when read are migrated to CalculationMethod. For version IMAER 5.1 calculation method will be written in the GML. For version IMAER 5.0 will be written in the GML.
In Calculation Method the method RADIUS has been removed as it's not used anymore.

TODO: Update the xsd to support calculation method. It will be an addition. (Although 'type' changes that it isn't required anymore) In this change the test gml files are already adapted to use 'method' element.

Possible TODO: rename PERMIT to FORMAL to align it better with the general usage and UK. (The code already contains some specific conversion in case FORMAL will be introduced).

This is a breaking change in the Java library.


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLCalculationSetOptionsReader.java
```diff
@@ -68,8 +69,21 @@ public CalculationSetOptions readCalculationSetOptions(final Theme theme) {
...
+   * Set the calculation method. If calculation method is null try if calculation type was used.
+   *
+   * @param calculationMetaData
+   * @param options
+   */
+  private void setCalculationMethod(final IsCalculationMetaData calculationMetaData, final CalculationSetOptions options) {
+    final String method = calculationMetaData.getCalculationMethod();
+
+    options.setCalculationMethod(
+        method == null ? CalculationType.toCalculationMethod(method) : CalculationMethod.safeValueOf(calculationMetaData.getCalculationMethod()));
```
BertScholten, line 87:
```suggestion
        method == null ? CalculationType.toCalculationMethod(calculationMetaData.getCalculationType()) : CalculationMethod.safeValueOf(method));
```
Not entirely sure if this is correct, but as it is it doesn't seem right either.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v5_0/togml/GMLVersionWriterV50.java
```diff
@@ -181,6 +181,19 @@ private CalculationMetadata getCalculation(final MetaDataInput input) {
...
+    final CalculationMethod calculationMethod = input.getOptions().getCalculationMethod();
+
+    if (calculationMethod == null) {
+      return null;
+    }
+    if (calculationMethod == CalculationMethod.PERMIT) {
+      return CalculationMethod.PERMIT.type();
+    } else {
+      return calculationMethod.type();
+    }
```
BertScholten, line 194: Not sure why this if/else is needed. Wouldn't just returning `calculationMethod.type()` do the same thing?

Also, I think we can get rid of the whole `v5_0/togml/` package now (perhaps different PR).

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/GMLCalculationSetOptionsReaderTest.java
```diff
@@ -76,7 +78,7 @@ void testReadCalculationSetOptionsNullCalculationOptions() {
     final FeatureCollection featureCollection = mock(FeatureCollection.class);
     final MetaData metaData = mock(MetaData.class);
     final IsCalculationMetaData calculationMetaData = mock(IsCalculationMetaData.class);
-    when(calculationMetaData.getCalculationType()).thenReturn("PERMIT");
+    when(calculationMetaData.getCalculationType()).thenReturn(CALCULATION_METHOD_FORMAL);
```
BertScholten, line 81: Would have expected this method to mock `getCalculationMethod()` instead with the rename, and perhaps another test method to check that it falls back to `getCalculationType()`.

BertScholten (changes requested):

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLCalculationSetOptionsReader.java
Hilbrand: Should be as you suggest.

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/GMLCalculationSetOptionsReaderTest.java
Hilbrand: Yes should use new method.

BertScholten (approved):
LGTM (though added some changes of my own to fix the build)

PR #182, opened 2023-02-22
AER-1348 add new calculation job type enum

Gijsvanhorn:


JornC (approved):

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationJobType.java
```diff
@@ -0,0 +1,126 @@
...
+ */
+public enum CalculationJobType {
+
+  /**
+   * Exactly 1 Project scenario
+   * Optional: 1 off site reduction
+   * Optional: 1 Reference
+   * NO other scenario types
+   */
+  PROCESS_CONTRIBUTION(new HashSet<>(Arrays.asList(SituationType.PROPOSED)),
```
BertScholten, line 40: `EnumSet` doesn't work?

Gijsvanhorn: I thought `EnumSet` isn't supported in GWT

PR #183, opened 2023-02-22
AER-1377 calculation job type in imaer

BertScholten:
Added the new calculation job type to CalculationMetaData in 5.1 (optional, any text, and will be coerced into enum when reading).


Gijsvanhorn (approved):
LGTM

PR #186, opened 2023-03-01
AER-1379 add single scenario type

Gijsvanhorn:
and add min and max situation validation information

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationJobType.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationJobType.java
@@ -37,7 +37,8 @@ public enum CalculationJobType {
    * Optional: 1 Reference
    * NO other scenario types
    */
-  PROCESS_CONTRIBUTION(new HashSet<>(Arrays.asList(SituationType.PROPOSED)),
+  PROCESS_CONTRIBUTION(
+      new HashSet<>(Arrays.asList(SituationType.PROPOSED)),
       new HashSet<>(Arrays.asList(SituationType.REFERENCE, SituationType.NETTING)),
       new HashSet<>()),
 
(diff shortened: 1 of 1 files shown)


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationJobType.java
```diff
@@ -59,24 +63,41 @@ public enum CalculationJobType {
...
+  private final int maximumNumberOfSituations;
+
   CalculationJobType(final Set<SituationType> requiredSituationTypes, final Set<SituationType> optionalSituationTypes,
-      final Set<SituationType> multipleSituationsAllowed) {
+      final Set<SituationType> multipleSituationsAllowed, final int minimumNumberOfSituations, final int maximumNumberOfSituations) {
     this.requiredSituationTypes = requiredSituationTypes;
     this.optionalSituationTypes = optionalSituationTypes;
     this.pluralSituationTypes = multipleSituationsAllowed;
+    this.minimumNumberOfSituations = minimumNumberOfSituations;
+    this.maximumNumberOfSituations = maximumNumberOfSituations;
```
BertScholten, line 100: Not sure what the point of these are, and if it makes sense to keep them in this enum: minimum is always 1, and maximum isn't as strict for things like `IN_COMBINATION_PROCESS_CONTRIBUTION` is it? Isn't it usually a situation type that should be at most 1 in a calculation job?

Hilbrand (changes requested):
This is not the right approach. The limit of 6 is also just an arbitrary limit made for the user interface, which should no tbe hardcoded in this library. I suggest to close this pr and add implement the checks in connect server. Put them in a separate validation class that checks for these limits.

Gijsvanhorn:
I removed the number of situations from the enum, and am doubting whether to remove the other validation rules as well. This PR is still relevant because it adds the SINGLE_SITUATION job type though.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationJobType.java
```diff
@@ -63,10 +65,19 @@ public enum CalculationJobType {
 
   /**
    * Only 1 Reference scenario
-   *
+   * <p>
```
BertScholten, line 68: I don't really mind this, but we don't use these tags that often and it does seem inconsistent with the other javadocs in this enum.

BertScholten (approved):
LGTM

Hilbrand (approved):
LGTM. Other validations can be left in.They specify something about the CalculationJobType.

PR #188, opened 2023-03-13
AER-1314 Add ROADS to custom OPS options

wtenbosch:


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/OPSOptions.java
```diff
@@ -140,6 +141,17 @@ public void setConvRate(final String convRate) {
     this.convRate = convRate;
   }
 
+  /**
+   * OPS .ctr variable: ROADS
+   */
```
Hilbrand, line 146: Maybe add javadoc (If I'm correct): `Space separated list of OPS cat. codes. That when match with a cat. code in the source cat column mean the source is interpreted by OPS as a road source.`

Hilbrand (approved):
LGTM

PR #194, opened 2023-04-07
Added Spatially Varying Roughness ADMS option

Hilbrand:
Also always write ADMS options affected when using demo mode to the GML (depletion, spatially variying roughness and, complex terrain)


BertScholten (approved):
LGTM

PR #200, opened 2023-05-04
AER-1619 heat content (and heat capacity) validation

BertScholten:
We didn't validate characteristics in this generic part yet (only on model objects, which is a bit limited and late in the process), so had to add some new validators.

Tbh the heat capacity isn't directly part of the issue, so that can be considered a bonus (felt wrong to just do something for OPS and leave ADMS hanging).

### DIFF
--- a/source/imaer-util/src/test/java/nl/overheid/aerius/validation/OPSCharacteristicsValidatorTest.java
+++ b/source/imaer-util/src/test/java/nl/overheid/aerius/validation/OPSCharacteristicsValidatorTest.java
@@ -31,18 +31,15 @@ import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
 import nl.overheid.aerius.shared.domain.ops.OPSLimits;
-import nl.overheid.aerius.shared.domain.v2.characteristics.ADMSSourceCharacteristics;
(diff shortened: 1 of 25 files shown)


source/imaer-util/src/test/java/nl/overheid/aerius/validation/OPSCharacteristicsValidatorTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+    final List<AeriusException> errors = new ArrayList<>();
+    final List<AeriusException> warnings = new ArrayList<>();
+    final OPSCharacteristicsValidator validator = new OPSCharacteristicsValidator(errors, warnings, SOURCE_ID);
+
+    final boolean valid = validator.validate(characteristics);
+
+    assertTrue(valid, "Valid test case");
+    assertTrue(errors.isEmpty(), "No errors");
+    assertTrue(warnings.isEmpty(), "No warnings");
```
Hilbrand, line 60: Maybe compare to an empty `List.of()`. That way if the list is not empty the error values are shown in the test output. With only test on empty you won't see what the list contains if it's not empty.

Hilbrand (commented):
Comment about test against empty list in asserts instead of test isEmpty to be able to see actual content if not empty. assert is used on multiple places, not only on commented place.

Hilbrand (approved):
LGTM

PR #210, opened 2023-06-20
AER-1696 AER-1833 IMAER UK changes

BertScholten:
- Adding gmlCreator to metadata
- Adding height and assessmentCategory to calculation point
- Switching to IMAER 5.1.2 (and as a result the artifact version to 5.1.2-SNAPSHOT)


wtenbosch (approved):
LGTM

PR #212, opened 2023-07-20
AER-1821 For NCA Calculation points make sure height is always saved in the GML

Hilbrand:


BertScholten (approved):
LGTM (though guess the height should be set somewhere before calculating as well?)

Hilbrand:
Height was already set when starting a calculation. But this is part of the code that converts object to ADMS specific objects (in Calculator/chunker)

PR #217, opened 2023-07-26
Adjusted farmland to not export same property twice

BertScholten:
As it was, if jackson's objectmapper was not explicitly set to exclude null properties, it would export activityType twice. While this is not really a problem when using the correct settings, it's cleaner to not have a property that we shouldn't have anyway.


Gijsvanhorn (approved):
LGTM

PR #218, opened 2023-07-28
AER-1006

Gijsvanhorn:
WIP because:

- ~~I've manually adjusted the XSD, so changes are not included in the EAP file. Also it is just a copy of `NETTING` now~~ :sweat_smile: 
- ~~I feel like handling the deprecated `NETTING` in the way I do now is not the right path, and we probably want a warning to be added about the use of a deprecated enum value.~~


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLValidator.java
```diff
@@ -47,6 +49,13 @@ public static void validateYear(final int year, final List<AeriusException> exce
     }
   }
 
+  public static void validateSituationType(final SituationType type, final ImportParcel parcel) {
+    if(type == SituationType.NETTING) {
+      parcel.getWarnings().add(new AeriusException(ImaerExceptionReason.GML_DEPRECATED_NETTING_UPDATED_TO_OFF_SITE_REDUCTION,
+          "Deprecated situation type NETTING will be updated to OFF_SITE_REDUCTION."));
+    }
+  }
```
BertScholten, line 57: I think we should just convert `NETTING` to `OFF_SITE_REDUCTION` on import, and make sure that we don't return any GMLs containing `NETTING` anymore. For old GMLs we'll always have to support a `NETTING` value anyway.

When we actually remove the deprecated `NETTING` type in IMAER we probably need a separate enum containing removed types for older GML versions. Perhaps we could introduce that one now, and only use `OFF_SITE_REDUCTION` in our domain objects?

(also, missing a space after the `if`)

BertScholten (changes requested):

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLValidator.java
Gijsvanhorn: I've interpreted your comment as introducing a new enum for all situationtypes including netting, removing `NETTING` from `SituationType`, and mapping the `LegacySituationType` to `SituationType` in `GMLReader`. (see my last commit). If you had something else in mind, please let me know because I'm not sure this is much better :sweat_smile:

source/imaer-gml/src/main/java/nl/overheid/aerius/importer/ImaerImporter.java
```diff
@@ -150,10 +150,9 @@ public void importStream(final InputStream inputStream, final Set<ImportOption>
   private static ScenarioSituation addSituationProperties(final GMLReader reader, final ImportParcel result) {
     final ScenarioSituation situation = result.getSituation();
     situation.setName(reader.getName());
-    final SituationType type = reader.getSituationType();
-    situation.setType(type);
+    situation.setType(reader.getSituationType());
     situation.setReference(reader.metaDataReader().readReference());
-    if (type == SituationType.NETTING) {
+    if (situation.getType() == SituationType.OFF_SITE_REDUCTION) {
```
BertScholten, line 155: Bit curious about this one, why was the local variable removed? Setters/getters normally don't affect performance I guess so it shouldn't matter, but wonder if there's another reason to do it like this?

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/GMLRoundtripTest.java
```diff
@@ -248,6 +247,19 @@ private ImportParcel importAndCompare(final String versionString, final String f
...
+      result.getSituation().setType(SituationType.PROPOSED);
+    }
+  }
+
+  private static String revertAutoCorrections(final String gml, final AeriusGMLVersion targetVersion) {
+    if (targetVersion.ordinal() >= AeriusGMLVersion.V5_1.ordinal()) {
+      return gml.replaceAll(SituationType.OFF_SITE_REDUCTION.name(), LegacySituationType.NETTING.name());
+    }
+    return gml;
+  }
```
BertScholten, line 261: Is this still needed?

source/imaer-gml/src/test/resources/gml/latest/validate/errors/fout_5239_netting_deprecated.gml
BertScholten, line 1: Don't think this error code exists anymore in this PR

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLValidator.java
BertScholten: Yes, I think it's better, now there's no chance we'll be using `NETTING` anywhere in non-GML code.

BertScholten (commented):
Some questions/suggestions

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/metadata/LegacySituationType.java
```diff
@@ -0,0 +1,75 @@
...
+   */
+  COMBINATION_REFERENCE,
+  /**
+   * Represents a situation that is to be calculated in cumulation with the proposed situation.
+   */
+  COMBINATION_PROPOSED,
+  /**
+   * Baseline situation
+   */
+  BASELINE;
```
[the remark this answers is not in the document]
Gijsvanhorn: true! I'll map `LegacySituationType.BASELINE` to `SituationType.UNKNOWN`

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/GMLRoundtripTest.java
Gijsvanhorn: unfortunately yes because the test gmls contain `NETTING` but after reading and writing this will now always become `OFF_SITE_REDUCTION` so the roundtrip is not completely the same. I've build in this check on the version assuming that gml version 5.2 and up will never contain `NETTING` as situationtype.

source/imaer-gml/src/main/java/nl/overheid/aerius/importer/ImaerImporter.java
Gijsvanhorn: not really, this change was left behind after some changes were made and partially reverted in this file. I've restored the local variable.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/FeatureCollection.java
```diff
@@ -18,6 +18,7 @@
 
 import java.util.List;
 
+import nl.overheid.aerius.gml.base.metadata.LegacySituationType;
```
BertScholten, line 21: Unnecessary import

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/MetaData.java
```diff
@@ -16,6 +16,7 @@
  */
 package nl.overheid.aerius.gml.base;
 
+import nl.overheid.aerius.gml.base.metadata.LegacySituationType;
```
BertScholten, line 19: Unnecessary import

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/metadata/LegacySituationType.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * Returns the {@link LegacySituationType} from the given string or {@link #UNKNOWN} if null or invalid input.
+   * @param type type to parse
+   * @return situation type object
+   */
+  public static LegacySituationType safeValueOf(final String type) {
+    try {
+      return type == null ? UNKNOWN : valueOf(type);
```
BertScholten, line 83: I'm not sure if it works to return `UNKNOWN` in all cases: old GMLs did not have the situation type, so think we always expect `null` for those. Unless you changed that in AERIUS-II repo as well?

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v5_0/togml/GMLVersionWriterV50.java
```diff
@@ -120,7 +121,7 @@ private SituationMetadata getSituation(final MetaDataInput input) throws AeriusE
       situation = new SituationMetadata();
       situation.setName(input.getName());
       situation.setReference(input.getReference());
-      situation.setSituationType(input.getSituationType());
+      situation.setSituationType(input.getSituationType() != null ? LegacySituationType.safeValueOf(input.getSituationType().name()) : null);
```
BertScholten, line 124: Hah, this is a good one. In this case I guess it should write `NETTING` instead of `OFF_SITE_REDUCTION` to still fit the 5.0 XSD?

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v5_1/togml/GMLVersionWriterV51.java
```diff
@@ -122,7 +123,7 @@ private SituationMetadata getSituation(final MetaDataInput input) throws AeriusE
       situation = new SituationMetadata();
       situation.setName(input.getName());
       situation.setReference(input.getReference());
-      situation.setSituationType(input.getSituationType());
+      situation.setSituationType(input.getSituationType() != null ? LegacySituationType.safeValueOf(input.getSituationType().name()) : null);
```
BertScholten, line 126: When writing 5.1, I don't think we should use `LegacySituationType` at all.

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/GMLRoundtripTest.java
BertScholten: Then I wouldn't mind if 5.1 roundtrip GML already contained `OFF_SITE_REDUCTION` instead. That way you don't have to do these auto corrections I think? When reading older versions it should be converted, and when writing that one again it should contain the new value.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/metadata/LegacySituationType.java
Gijsvanhorn: I think it will work, the original SituationType also mapped `null` to `UNKNOWN` so I don't see a difference there

BertScholten: Ah yeah, guess you're right there.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v4_0/metadata/MetaDataImpl.java
```diff
@@ -127,7 +127,7 @@ public String getSituationName() {
 
   @Override
   public SituationType getSituationType() {
-    return situation == null ? null : situation.getSituationType();
+    return situation.getSituationType().toSituationType();
```
BertScholten, line 130: Might want to leave the null checks in here, SituationMetaData is not a required field.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v5_1/togml/GMLVersionWriterV51.java
Gijsvanhorn: agreed

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v4_0/metadata/MetaDataImpl.java
```diff
@@ -127,7 +127,7 @@ public String getSituationName() {
 
   @Override
   public SituationType getSituationType() {
-    return situation == null ? null : situation.getSituationType();
+    return situation.getSituationType() != null ? situation.getSituationType().toSituationType() : null;
```
BertScholten, line 130:
```suggestion
    return situation == null || situation.getSituationType() == null ? null : situation.getSituationType()toSituationType();
```

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v5_0/metadata/MetaDataImpl.java
```diff
@@ -128,7 +128,7 @@ public String getSituationName() {
 
   @Override
   public SituationType getSituationType() {
-    return situation == null ? null : situation.getSituationType();
+    return situation.getSituationType() != null ? situation.getSituationType().toSituationType() : null;
```
BertScholten, line 131: see other metadataimpl ;)

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v5_1/metadata/MetaDataImpl.java
```diff
@@ -139,7 +139,7 @@ public String getSituationName() {
 
   @Override
   public SituationType getSituationType() {
-    return situation == null ? null : situation.getSituationType();
+    return situation.getSituationType();
```
BertScholten, line 142: null check (change can be reverted)

BertScholten (approved):
LGTM

PR #219, opened 2023-08-08
AER-1006 Allow netting in 5.1 

Gijsvanhorn:
the situationtype `NETTING` is valid in version 5.1 according to its xsd (and can now be exported from production). But it could not be read, and would be defaulted back to PROPOSED.


BertScholten (approved):
LGTM

PR #223, opened 2023-10-11
AER-2206 archive imaer changes

BertScholten:
Switched to IMAER 5.1.3, so updated versions accordingly.
Included a roundtriptest as a showcase of how an Archive GML could look like (not too sure how much project information and other metadata should be included, this seems the bare minimum).


Hilbrand (commented):
LGTM. Can you also add the `CalculationSetOptions` field `useInCombinationArchive` as added in ui and add it to `CalculationMetadata` or is that something for `OptionsMetadataUtil`?

PR #225, opened 2023-10-18
Ensuring ID for subpoints includes both subpoint ID and receptor ID

BertScholten:
As this combination is going to be unique, it makes sense to have these IDs be combined in the default GML ID(s). Added a second point with the same subpoint ID to the existing test cases: these would fail some tests in the old situation.


Hilbrand (approved):
LGTM

PR #226, opened 2023-10-18
Excluding some packages from GWT

BertScholten:
ArchiveMetaData contains ZonedDateTime, and that is not emulated by GWT apparantly. To avoid having to resort to some other way (string perhaps?), we might as well exclude some packages, as these shouldn't be serialized to GWT code anyway.


Hilbrand:
The fact it doesn't compile for a shared class kind of breaks the contract of the shared project. The shared project is intended to be compatible with GWT. Can't we use `Date`  in the v2 objects?

BertScholten:
I guess we could use `Date`, but rather avoid that if we don't have to (we'd have to convert from one to the other at some point). What do you think of this latest approach?

Hilbrand (approved):
Lets keep this solution for now.

PR #227, opened 2023-11-14
AER-2449 constructArchiveMetaData does not set ScenarioMetaData

Hilbrand:
Therefore we should check if scenarioMetaData is not null as it's not required to have this information set.


BertScholten (approved):
LGTM

PR #236, opened 2024-01-04
AER-1036 log validation errors as debug since message is rethrown again and later logged if needed.

Hilbrand:
Also some qa improvements included.


BertScholten (approved):
LGTM

PR #237, opened 2024-01-12
Updated gwt-maven-plugin, removed gwt version numbers

Hilbrand:
These gwt version numbers are not used in the pom.xml files and therefor can be removed.


BertScholten (approved):
LGTM

PR #238, opened 2024-02-02
AER-2656 Added Monthly time varying emission profile to enum

Hilbrand:


BertScholten (approved):
LGTM

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/characteristics/CustomDiurnalVariationType.java
```diff
@@ -43,7 +43,12 @@ public double sum(final List<Double> values) {
       return sumWeekdays + sumOtherDays;
     }
 
-  };
+  },
+
+  /**
+   * Monthly Time-varying emission profile over a year, with 1 value for each month.
+   */
+  MONTHLY(12);
```
BertScholten, line 51: Could be `MONTH` to be more in lin with `DAY`?

Hilbrand: I followed ADMS manual. This uses `3-day` and `monthly`. Therefore I guess monthly seems better.

BertScholten: In that case let's follow the native speakers I guess.

PR #240, opened 2024-02-05
AER-2477 Validate by schema when writing GML

BertScholten:
By setting the schema on the marshaller, any XSD violations should be caught before invalid GMLs end up at the user.

This will cause some unittests to fail upstream, but that's for later concern.

### DIFF
--- a/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/GMLSchemaFactory.java
+++ b/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/GMLSchemaFactory.java
@@ -17,8 +17,6 @@
 package nl.overheid.aerius.gml.base;
 
 import java.net.URL;
-import java.util.EnumMap;
-import java.util.Map;
 
 import javax.xml.XMLConstants;
 import javax.xml.validation.Schema;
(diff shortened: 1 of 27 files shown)


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/GMLSchemaFactory.java
```diff
@@ -0,0 +1,85 @@
...
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V1_0, nl.overheid.aerius.gml.v1_0.base.CalculatorSchema.SCHEMA_LOCATION);
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V1_1, nl.overheid.aerius.gml.v1_1.base.CalculatorSchema.SCHEMA_LOCATION);
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V2_0, nl.overheid.aerius.gml.v2_0.base.CalculatorSchema.SCHEMA_LOCATION);
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V2_1, nl.overheid.aerius.gml.v2_1.base.CalculatorSchema.SCHEMA_LOCATION);
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V2_2, nl.overheid.aerius.gml.v2_2.base.CalculatorSchema.SCHEMA_LOCATION);
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V3_0, nl.overheid.aerius.gml.v3_0.base.CalculatorSchema.SCHEMA_LOCATION);
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V3_1, nl.overheid.aerius.gml.v3_1.base.CalculatorSchema.SCHEMA_LOCATION);
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V4_0, nl.overheid.aerius.gml.v4_0.base.CalculatorSchema.SCHEMA_LOCATION);
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V5_0, nl.overheid.aerius.gml.v5_0.base.CalculatorSchema.SCHEMA_LOCATION);
+    SCHEMA_LOCATIONS.put(AeriusGMLVersion.V5_1, nl.overheid.aerius.gml.v5_1.base.CalculatorSchema.SCHEMA_LOCATION);
```
Hilbrand, line 55: I guess we could put the `CalculatorSchema` data in the `AeriusGMLVersion` enum as that  information is consistent over all versions. That way we don't have to update it here (or forget it) when the a new version is added.

Hilbrand (approved):
LGTM. I was more thinking of moving all content of CalculatorSchema and put all methods on enum and remove the CalculatorSchema classes as all have the same content.

BertScholten:
@Hilbrand oh like that. I don't think we can do that with all the content as some of those properties are referenced from within the jaxb annotations (`CalculatorSchema.NAMESPACE`) so that would cause some circular references. But the schema location might be something that we can move, that should only be a locally used field. I'll have a look tomorrow.

PR #241, opened 2024-02-07
AER-2658 Added monthly variation related fields in IMAER for ADMS

BertScholten:
This is an adjustment that is OK to add to the current IMAER version as it's only adding an optional field. It's also the easy solution where we just have a separate field for monthly-based profiles in ADMS characteristics. This also fits with what ADMS supports. While we could opt to use a 0..* field, and do the validation on correct type usage in code, this does have the drawback of supporting more types and being less intuitive to the user (I imagine) as well as not being as backwards-compatible.

Did put some notes in IMAER descriptions that a few objects are up for renaming. Might be wishful thinking, but I'd rather like a cleanup version for our next release.


Hilbrand (approved):
LGTM

PR #245, opened 2024-02-15
AER-2361 - Round in gml writer

JornC:


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/geo/Geometry2GML.java
```diff
@@ -124,16 +130,22 @@ private static AbstractRingPropertyType setCoordinates(final ObjectFactory facto
 
   /**
    * Set coordinates from wkt line string to gml LineStringType
```
BertScholten, line 132: Might as well remove this whole comment, it doesn't really make sense... (or update it to something that does ;))

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/base/geo/Geometry2GMLTest.java
```diff
@@ -0,0 +1,80 @@
...
+      public void setGmlPoint(final PointType gmlPoint) {
+        this.gmlPoint = gmlPoint;
+      }
+    };
+  }
+
+  static Stream<Arguments> pointProvider() {
+    return Stream.of(
+        Arguments.of("Rounding up", new Point(5.00015, -5.00015), "5.0", "-5.0"),
+        Arguments.of("Rounding down", new Point(5.00014, -5.00014), "5.0", "-5.0"),
```
BertScholten, line 59: ehhhh what is rounding up/rounding down in this case? Shouldn't it be `5.0015` and `5.0014` if you want to test that?

And if not, might be nice to add a case that does something like `1.23456789` getting rounded to `1.235`.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/geo/Geometry2GML.java
JornC: Oh, frankly I meant to remove the fluff in the other method instead of this one :sweat:

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/base/geo/Geometry2GMLTest.java
JornC:
```
        Arguments.of("Rounding up positive", new Point(1.0001, -0.9999), "1.0", "-1.0"),
        Arguments.of("Rounding down negative", new Point(-0.9999, 1.0001), "-1.0", "1.0"),
        Arguments.of("Edge case rounding up", new Point(0.9995, -0.9995), "1.0", "-0.999"),
        Arguments.of("Edge case rounding down", new Point(0.9994, -0.9994), "0.999", "-0.999"),
```

I've added them like this (hopefully also very readable)

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/geo/Geometry2GML.java
JornC: --- removed the parameter fluff in the other method and left this one untouched

BertScholten (approved):
LGTM (none of the existing roundtrip tests failed?)

BertScholten:
Oh turns out some tests did fail:

```
Error:  Tests run: 386, Failures: 84, Errors: 0, Skipped: 0, Time elapsed: 27.603 s <<< FAILURE! - in nl.overheid.aerius.gml.GMLRoundtripTest
```
Perhaps a good idea to set that action to fail on unittest failures

BertScholten:
Also, the more I think about this, the less I like it in this location. This should be handled much earlier in the process, preferably somewhere before doing actual calculations. As it is now, you might end up with slightly different results when importing/recalculating the GML again as the calculation was done with different coordinates (perhaps not in the case of OPS, but maybe for some other models).

JornC:
> Also, the more I think about this, the less I like it in this location. This should be handled much earlier in the process, preferably somewhere before doing actual calculations. As it is now, you might end up with slightly different results when importing/recalculating the GML again as the calculation was done with different coordinates (perhaps not in the case of OPS, but maybe for some other models).

Hmm yeah that makes sense. And it can't only be in the client/Calculator either because there are multiple clients (i.e. direct Connect), so directly after a start calculation call arrives in Connect?

JornC:
@BertScholten and would it make sense to validate the # of decimals here and throw an error in case there are multiple (such as resulting from double point precision rounding, which I don't think happens [in Connect] and would be a bug on our end), so as to be able to guarantee the rounding behaviour?

JornC:
@BertScholten Come to think of it, what would possibly make the most sense is to serialize to GML, read that GML, and use the result for any calculations :thinking:

BertScholten:
That might make sense, but also doesn't work for all cases: there is a way to supply source coordinates in WGS84 (through Connect api and brn files), and those should not be rounded to 3 decimals. And GML is not our only export format, we also export CSV stuff through Connect.

Currently we serialize scenario's to JSON before sending it on to the worker. I think that would be a good spot to fix any coordinates.

PR #246, opened 2024-02-21
Get enum by EPSG code

JornC:


source/shared-geo/src/main/java/nl/overheid/aerius/shared/geo/EPSG.java
```diff
@@ -58,6 +58,16 @@ public static EPSG getEnumBySrid(final int srid) {
     return null;
   }
 
+  public static EPSG getEnumByEpsg(final String epsgCode) {
+    try {
+      final int srid = Integer.parseInt(epsgCode.substring(EPSG_PREFIX.length()));
+      return getEnumBySrid(srid);
+    } catch (final NumberFormatException e) {
```
BertScholten, line 65: Guess you can also get an exception if supplied string is null or does not have enough length?

BertScholten (approved):
LGTM

PR #247, opened 2024-02-23
AER-2710 rename WNB to OWN2000

Hilbrand:
Deprecated Theme WNB to avoid breaking backward compatibility.

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationSetOptions.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationSetOptions.java
@@ -38,7 +38,7 @@ public class CalculationSetOptions implements Serializable {
   private boolean useInCombinationArchive;
   private final ArrayList<Substance> substances = new ArrayList<>();
   private final Set<EmissionResultKey> emissionResultKeys = new HashSet<>();
-  private WNBCalculationOptions wnbCalculationOptions = new WNBCalculationOptions();
+  private OwN2000CalculationOptions own2000CalculationOptions = new WNBCalculationOptions();
   private RBLCalculationOptions rblCalculationOptions = new RBLCalculationOptions();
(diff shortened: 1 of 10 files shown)


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationSetOptions.java
```diff
@@ -38,7 +38,7 @@ public class CalculationSetOptions implements Serializable {
   private boolean useInCombinationArchive;
   private final ArrayList<Substance> substances = new ArrayList<>();
   private final Set<EmissionResultKey> emissionResultKeys = new HashSet<>();
-  private WNBCalculationOptions wnbCalculationOptions = new WNBCalculationOptions();
+  private OwN2000CalculationOptions own2000CalculationOptions = new WNBCalculationOptions();
```
SerhatG, line 41:
```suggestion
  private OwN2000CalculationOptions own2000CalculationOptions = new OwN2000CalculationOptions();
```

BertScholten: Think it's required for the cast in `getWnbCalculationOptions`. Will probably require a cleanup PR to change this once other repository is up to date.

BertScholten (approved):
LGTM

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/CalculationSetOptions.java
Hilbrand: Yes that is required because otherwise the old `getWnbCalculationOptions` won't work. When the pr in Calculator is merged I'll do a followup pr to clean up the deprecated code here. It might be an idea to do this when we change the IMAER-java version pom version, so it's a bit more separated.

PR #252, opened 2024-03-12
AER-2750 Shapechange 3.0.0 and cleanup

BertScholten:
Moved from ShapeChange 2.0.1 (which we have been using from the start) to 3.0.0 (which was released last year). Did mean all the documenation in the EAP file had to be updated to remove `-- Definition --` and `-- Description --` as those are no longer treated as a special case, but that was not really adding anything anyway. Also got rid of HTML that was added to version control, but which is actually generated during shapechange run. Also updated instructions and changed that to markdown.

Also renamed diurnal variation to time varying profiles (it was a note, wanted to get rid of those), and removed some deprecated and unused fields.

Also removed plan emission sources, which will require some updates in projects that use this library.


JornC (approved):
This PR removes a bunch of things but adds 10k lines to the repo :monocle_face::monocle_face: :monocle_face: :monocle_face: ACK

(edit: ah I see the 15k eap xmi document now)

PR #254, opened 2024-03-18
AER-2774 Added Cold Start emission source object and emission calculation

Hilbrand:
Also switch to Java 17 (except for `-shared` projects, which are used in GWT code).


BertScholten (approved):
LGTM

PR #255, opened 2024-03-22
AER-2772 - Split up cold start types

JornC:
Since there is no other simple way to distinguish between these types on the client, we need to add another emission source type.

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/EmissionSourceType.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/EmissionSourceType.java
@@ -43,7 +43,7 @@ public enum EmissionSourceType {
   /**
    * Cold start parking emission values.
    */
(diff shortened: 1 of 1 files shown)


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/EmissionSourceType.java
```diff
@@ -41,9 +41,13 @@ public enum EmissionSourceType {
    */
   MEDIUM_COMBUSTION_PLANT(Names.MEDIUM_COMBUSTION_PLANT),
   /**
-   * Cold start emission values.
+   * Cold start parking emission values.
    */
-  COLD_START(Names.COLD_START),
+  COLD_START_PARKING(Names.COLD_START_PARKING),
```
BertScholten, line 46: think it's not just parking, but more a parking garage? The other case is usually also parking, but then outside (or at least no building effects)

Hilbrand (approved):
LGTM

PR #256, opened 2024-03-25
AER-2554 cold start in IMAER

BertScholten:
Including roundtrip test case.

Also changed name of custom vehicle property from `emission` to `emissionFactor`, seemed more consistent/applicable.


source/imaer-gml/src/test/resources/gml/v6_0/roundtrip/coldstart_without_characteristics.gml
```diff
@@ -0,0 +1,112 @@
...
+            <imaer:version>
+                <imaer:VersionMetadata>
+                    <imaer:aeriusVersion>DEV</imaer:aeriusVersion>
+                    <imaer:databaseVersion>DEV_DB</imaer:databaseVersion>
+                </imaer:VersionMetadata>
+            </imaer:version>
+        </imaer:AeriusCalculatorMetadata>
+    </imaer:metadata>
+    <imaer:featureMember>
+        <imaer:ColdStartEmissionSource sectorId="3150" gml:id="ES.1">
```
Hilbrand, line 27: Shouldn't this be sectorId `3160`?

Hilbrand (changes requested):
The difference between cold start parking garage and others is now not visible from the data objects. Maybe we should add/change something to make this more explicit?

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v0_5/source/road/CustomVehicle.java
```diff
@@ -31,11 +31,11 @@
  *
  */
 @XmlRootElement(name = "CustomVehicle", namespace = CalculatorSchema.NAMESPACE)
-@XmlType(name = "CustomVehicleType", namespace = CalculatorSchema.NAMESPACE, propOrder = {"description", "emissions"})
+@XmlType(name = "CustomVehicleType", namespace = CalculatorSchema.NAMESPACE, propOrder = {"description", "emissionFactors"})
```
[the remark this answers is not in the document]
BertScholten: Proporder looks at the name of the property (getter/setter), and as I updated the (overriden) getter from `getEmissions` to `getEmissionFactors`, I also had to change something in this proporder. It will still expect `emission` as the element name for these old versions, that is defined in the getter annotation (and that should be covered by roundtrip tests).

source/imaer-gml/src/test/resources/gml/v6_0/roundtrip/coldstart_without_characteristics.gml
BertScholten: Maybe, the value doesn't matter at the moment though, as it is not hardcoded in the IMAER part for any decision.

The implementation lets the user free in specifying source characteristics or not. In the case of most sources, this means that characteristics in GML will be used if specified, and the sector default will be used if it wasn't specified. There are a few source types where characteristics will just be ignored if supplied, no matter if it was specified or not. ColdStart is new in the sense that characteristics will be read if present, but if they're not there then we won't fall back to sector defaults.

We could add another flag to cold start specifically, which would tell us if characteristics should be user defined or not, but not specifying it implicitly states the same?

BertScholten: As discussed, added a flag on the emission source level.

Hilbrand (approved):
LGTM

PR #257, opened 2024-04-04
[AER-2774] Changes related to cold start emission calculations.

Hilbrand:
AER-2846 - added vehicle type to CustomVehicle, that is needed to determine characteristics for cold start vehicles.

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/ColdStartEmissionsCalculator.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/ColdStartEmissionsCalculator.java
@@ -41,7 +41,7 @@ import nl.overheid.aerius.shared.exception.AeriusException;
  * 1000: divide by 1000 to convert gram to kilogram
  * </pre>
  */
-public class ColdStartEmissionsCalculator extends BaseRoadEmissionsCalculator<ColdStartEmissionSource> {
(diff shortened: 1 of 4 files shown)


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/ColdStartEmissionsCalculator.java
```diff
@@ -41,7 +41,7 @@
  * 1000: divide by 1000 to convert gram to kilogram
  * </pre>
  */
-class ColdStartEmissionsCalculator extends BaseRoadEmissionsCalculator<ColdStartEmissionSource> {
+public class ColdStartEmissionsCalculator extends BaseRoadEmissionsCalculator<ColdStartEmissionSource> {
```
BertScholten, line 44: Not entirely sure why this was made public?

BertScholten (approved):
LGTM

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/emissions/ColdStartEmissionsCalculator.java
Hilbrand: Looks like a remnant of an earlier development phase. It's not needed. So I've updated pr and reverted.

PR #258, opened 2024-04-10
AER-2860 Added optional vehicle type attribute to custom vehicle

BertScholten:
This should be filled in for cold start with vehicle based characteristics, otherwise it'll just be ignored on import. Also added this to validation, and added a test class for the validator.


JornC (approved):
ACK

PR #259, opened 2024-04-10
AER-2554 Fix serialization to/from json for cold start stuff

BertScholten:
Using 2 different names for the same source class is problematic. Reading is fine, but writing will always use 1 of the names (could maybe do something with another subtype defintion in a class between EmissionSource and ColdStartEmissionSource and then looking at the vehicleBasedCharacteristics value, but that's rather ugly and not even sure if that'd work). Better to use 1 name. Deprecated the existing values while we fix this, and added a new one that should be used instead. 

Also using a json annotation to specify a different name for vehicle type: vehicleType is already used for the distinction between standard, specific and custom, so can't use that again. 

Also added 2 json round trip cases for cold start.


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/EmissionSourceType.java
```diff
@@ -43,10 +43,20 @@ public enum EmissionSourceType {
   /**
    * Cold start parking emission values.
    */
+  COLD_START(Names.COLD_START_PARKING_GARAGE),
```
JornC, line 46:
```suggestion
  COLD_START(Names.COLD_START),
```

JornC (approved):
ACK

PR #261, opened 2024-04-23
AER-2827 animal housing domain, validation and emission calculation

BertScholten:
Deprecated a number of farm lodging related classes, as those are up for removal with introduction with the animal housing approach (which is similar, but quite a bit simpler).

Added some methods with a default implementation so this should not affect Calculator main code. 
(Locally it seems to work at least...)


JornC (approved):
ACK

PR #264, opened 2024-05-01
AER-2828 farm animal housing in IMAER

BertScholten:
Replaced FarmLodging with the FarmAnimalHousing approach in IMAER 6.0. Lodging in old versions is still supported, and 5.1 will still export it as such (for now at least).

Does not include backwards compatibility, that'll be another PR.

Might break some (integration) tests in upstream projects.


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v6_0/togml/Source2GML.java
```diff
@@ -164,14 +164,14 @@ public nl.overheid.aerius.gml.v6_0.source.EmissionSource visit(final OffRoadMobi
   @Override
   public nl.overheid.aerius.gml.v6_0.source.EmissionSource visit(final FarmLodgingEmissionSource emissionSource, final IsFeature feature)
       throws AeriusException {
-    return new Farm2GML().convert(emissionSource);
+    // Not supported from 6.0 onwards, just return a generic emission source.
+    return new nl.overheid.aerius.gml.v6_0.source.EmissionSource();
```
Hilbrand, line 168: I think this case should not be allowed and throw an exception? But I guess that would break the, among other things, round trip test? I would assume old farm lodging sources are converted to new farm housing sources. (Is that already supported?) I which case the round trip reference data for the old gml would contain the new housing data. I guess not all data can be converted, but for the tests would write them such that they are work.
Would it also make sense to give `EmissionSourceVisitor` a default implementation for `FarmLodgingEmissionSource` that throws the exception?

BertScholten: Backwards compatibility that will convert old farm lodging sources to the new farm animal housing approach is not yet implemented. Once it is, the round trip data should be updated accordingly. And then we can think about throwing an exception here.

Doing it like this is mostly for the case where on the main branch we're using farm lodging still, while we're implementing animal housing in a feature branch. That'll also show what (cypress) test cases are affected. We could switch the main branch to 5.1 exports again, but then cold start (once merged) will not properly export. Just exporting it as a generic source seems the lesser evil in this case.

Hilbrand (approved):
LGTM

PR #266, opened 2024-05-06
AER-1815 Added separate minimum height for source volumes, and changed defaults

Hilbrand:
Volume source default should be 0.0005 and other sources default should be 0.0.


BertScholten (approved):
LGTM

PR #267, opened 2024-05-13
AER-2858 farm lodging backwards compatibility

BertScholten:
This should not affect upstream projects: as long as the map with conversions is empty, the old conversion will take place and sources will be imported with the legacy classes. If the map is not empty, every lodging will be converted to animal housing.

This conversion is based on description/file in Jira, so only main lodging is converted to animal housing (with an optional extra additional system), while any additional (or reductive) systems or fodder measures on the lodging are ignored. Simple for implementation, though does mean we had to put in some specific warnings.


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/lodging/GML2Farm.java
```diff
@@ -37,16 +45,121 @@
...
+  private static final String UNKNOWN_ANIMAL_TYPE_CODE = "";
+
   public GML2Farm(final GMLConversionData conversionData) {
     super(conversionData);
   }
 
   @Override
-  public FarmLodgingEmissionSource convert(final T source) throws AeriusException {
+  public EmissionSource convert(final T source) throws AeriusException {
+    EmissionSource converted;
```
Hilbrand, line 60:
```suggestion
    final EmissionSource converted;
```

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/GMLRoundtripTest.java
```diff
@@ -191,8 +201,9 @@ private static void addArguments(final List<Object[]> files, final AeriusGMLVers
     }
   }
 
-  private static boolean skipThisTest(final AeriusGMLVersion version, final String file) {
-    boolean skipThisTest = false;
+  private static boolean skipThisTest(final AeriusGMLVersion version, final String file, final AeriusGMLVersion targetVersion,
+      final boolean skipOld) {
+    boolean skipThisTest = false || (skipOld && targetVersion != CURRENT_GML_VERSION);
```
Hilbrand, line 206:
```suggestion
    boolean skipThisTest = skipOld && targetVersion != CURRENT_GML_VERSION;
```

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/lodging/GML2Farm.java
```diff
@@ -37,16 +45,121 @@
...
+    if (getConversionData().hasFarmLodgingConversions()) {
+      converted = convertToAnimalHousing(source);
+    } else {
+      // Temporary while we implement this in a feature branch. Once upstream main is set, we can remove this along with the deprecated class.
+      converted = convertLegacy(source);
+    }
+    return converted;
+  }
+
+  public FarmAnimalHousingEmissionSource convertToAnimalHousing(final T source) throws AeriusException {
```
Hilbrand, line 70: Is this method public to use elsewhere?

BertScholten: Nope, changed it to private.

Hilbrand (approved):
LGTM

PR #268, opened 2024-05-22
Rename/Deprecate RBL for CIMLK

Hilbrand:
Put this naming change to avoid having future updates to IMAER-Java in case only change would require updating library only to have this rename.


BertScholten (approved):
LGTM

PR #270, opened 2024-05-23
AER-2918 Met site changes

BertScholten:
Added MetDatasetType enum, and properties in ADMSOptions to support the new met site selection (dataset type and years). Deprecated met site properties on NCACalculationOptions: we never really used these.


JornC (approved):
ACK

PR #271, opened 2024-05-24
AER-2979 Validating additional systems based on housing

BertScholten:
Data apparantly also contains additional systems that should only be used for one specific housing category. As that is more specific, it's better to always check on housing category over the animal category.


JornC (approved):
ACK

PR #277, opened 2024-05-30
AER-2799 Relaxed validation on allowed time varying profile values

BertScholten:
Now also allowing the exact limit instead of throwing an exception (similar to frontend in related project).


Hilbrand (approved):
LGTM

PR #278, opened 2024-05-30
Changed PR github action to fail when test fails

BertScholten:
Release already fails if this is the case, so makes more sense to fail when a PR introduces new failing tests. (noticed some tests failing, thought it already worked this way)


SerhatG:
I am okay with making this the default in `github-actions` as long as we are aware this is the case and it will impact all opensource libraries.

SerhatG (approved):

PR #284, opened 2024-06-11
AER-2978 Possibility to use warnings for building upper limits

BertScholten:
NL (OPS) does not want errors for upper limits, but wants warnings instead. Bit annoying, but this solution should work for that requirement. Also applied it to diameter, though NL does not really use that. Felt too weird to call it 'isBuildingHeightUpperLimitWarning' however, so went with this.


JornC (approved):
ACK

PR #290, opened 2024-07-09
AER-2977 Geometry validation for ADMS + max vertices

BertScholten:
That is, when characteristics dictate certain source types, the geometry should match it.
Also added a test on maximum vertices in the case of a polygon (area or surface).


Hilbrand (approved):
LGTM

PR #291, opened 2024-08-01
AER-3104 Import screen error formatting

MichielJanssen-DAT:
- Added unique error line seperator

### DIFF
--- a/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLReaderFactory.java
+++ b/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLReaderFactory.java
@@ -169,7 +169,7 @@ public final class GMLReaderFactory {
     final StringBuilder errorLine = new StringBuilder();
     for (final ValidationEvent event : list) {
       errorLine.append(event.getMessage());
-      errorLine.append(' ');
+      errorLine.append('|');
     }
     LOG.debug("validation error: {}", errorLine);
     return new AeriusException(ImaerExceptionReason.GML_VALIDATION_FAILED, errorLine.toString().trim());


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLReaderFactory.java
```diff
@@ -169,7 +169,7 @@ private static AeriusException newValidationFailed(final List<ValidationEvent> l
     final StringBuilder errorLine = new StringBuilder();
     for (final ValidationEvent event : list) {
       errorLine.append(event.getMessage());
-      errorLine.append(' ');
+      errorLine.append('|');
```
BertScholten, line 172: This is not how we want to do this: with this you have to do some parsing of the message where you want to show this to a user, and would like to avoid that.
Instead, each event should end up in its own exception, and this validation should return a list of exceptions.

BertScholten (changes requested):

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/GMLValidateErrorsTest.java
```diff
@@ -333,6 +343,14 @@ private static <T extends Throwable> T assertResult(final String fileName, final
         "Expected an exception " + clazz.getSimpleName());
   }
 
+  private static void assertResults(final String fileName, final List<String> expectedReasonsTxt, final Reason expectedReason) throws IOException, AeriusException {
+    final ImportParcel result = getImportResult(LATEST_VALIDATE, fileName);
+    for (int i = 0; i < expectedReasonsTxt.size(); i++) {
```
BertScholten, line 349: Might want to assert that the lists have equal size.

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/GMLValidateErrorsTest.java
```diff
@@ -333,6 +343,14 @@ private static <T extends Throwable> T assertResult(final String fileName, final
         "Expected an exception " + clazz.getSimpleName());
   }
 
+  private static void assertResults(final String fileName, final List<String> expectedReasonsTxt, final Reason expectedReason) throws IOException, AeriusException {
```
BertScholten: `expectedReason` is not used/checked.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLReaderFactory.java
```diff
@@ -116,12 +120,13 @@ public ValidationHelper createValidationHelper() {
     return gmlHelper.getValidationHelper();
   }
 
-  private static void checkEvents(final List<ValidationEvent> events) throws AeriusException {
+  private static List<AeriusException> checkEvents(final List<ValidationEvent> events) throws AeriusException {
     // information like required elements not being present will not throw an exception, but will trigger an event.
     // check for these events and throw exception if present.
     if ((events != null) && !events.isEmpty()) {
-      throw newValidationFailed(events);
+      return newValidationFailed(events);
```
BertScholten, line 127: comment above doesn't really match anymore

BertScholten (approved):
LGTM

PR #292, opened 2024-08-07
AER-3128 Animal housing less animal type based

BertScholten:
Basic housing is now based on animal housing category instead of the animal type. This is due to some animal types having multiple basic animal housing categories. Similarly, additional system factors are now based on animal housing category instead of the animal type. This is due to having different reduction factors for some animal housing + additional system combinations. Implementations will have to be adjusted accordingly.

### DIFF
--- a/source/imaer-gml/src/test/java/nl/overheid/aerius/test/TestValidationAndEmissionHelper.java
+++ b/source/imaer-gml/src/test/java/nl/overheid/aerius/test/TestValidationAndEmissionHelper.java
@@ -815,12 +815,12 @@ public class TestValidationAndEmissionHelper implements ValidationHelper, Emissi
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


source/imaer-gml/src/test/java/nl/overheid/aerius/test/TestValidationAndEmissionHelper.java
```diff
@@ -815,12 +815,12 @@ public boolean isAdditionalHousingSystemAirScrubber(final String additionalSyste
   }
 
   @Override
-  public String getAnimalBasicHousingCode(final String animalCode) {
-    return animalCode + ".100";
```
JornC, line 819: Hah..

JornC (approved):

PR #294, opened 2024-08-20
Extended existing test slightly for xsd errors

BertScholten:
For future changes, where we might want to filter out the first one, or group based on location. It's also the original test case.


Hilbrand (approved):
LGTM

PR #297, opened 2024-08-30
AER-3178 Fix cold start SubSourceEmissionsCalculator

Hilbrand:
The cold start emissions calculator should return the emissions in kg/Year (it was g/year).


BertScholten (approved):
LGTM, bit different than I thought at first glance, but can see why its needed. Other option might have been to introduce a 2nd `calculateColdStartEmissions` method within `ColdStartEmissionsCalculator` that does the conversion, and make the current one protected?

Hilbrand:
@BertScholten Yes that was my first approach to add that method. However, there are 2 method calls to different methods. Therefore it would require to add 2 methods, meaning 2 additional methods to the class. So I opted for the approach implemented, to keep number of methods smaller.

PR #301, opened 2024-09-09
AER-2804 Added option to not add receptor hexagon representation in results in GML

Hilbrand:
Change is backward compatible.


BertScholten (approved):
LGTM

PR #302, opened 2024-09-09
AER-2804 serialize ScenarionSituationResults to be ably to pass calculation id via message bus

Hilbrand:
Than the calculation id can be used to retrieve the results to put in the GML. But we don't want to serialize results in this class as those are not needed, but only add to the message load.


BertScholten (approved):
LGTM (I think, not too sure how it ends up being used)

PR #303, opened 2024-09-27
AER-3099 Added support for having multi adms met site properties for multiple years.

Hilbrand:
This code introduces a new MetSurfaceCharacteristics object that replaces the single values in ADMSOptions. For backward compatibility the individual methods are kept in ADMSOptions, and will be removed when not used any more by upstream project. When multiple years are present in the met site data it's expected to have multiple met values for these years. In the calculation options these will be saved with the year as prefix to the option. If only 1 (or no) year is present no prefix will be used.


BertScholten (approved):
LGTM 

Do wonder if we want to have met site characteristics be exported (not something user should have input on, and we shouldn't use the imported value?), but guess there was a reason for it at some point...

PR #304, opened 2024-10-10
AER-3210 Added ProjectCategory enum

MichielJanssen-DAT:
- Added ProjectCategory enum
- Added ProjectCategory to NCACalculationOptions
- Added ProjectCategory to OptionsMetadataUtil and -test


source/imaer-util/src/test/java/nl/overheid/aerius/util/OptionsMetadataUtilTest.java
```diff
@@ -198,6 +199,7 @@ void testNcaOptions() {
     final NCACalculationOptions ncaOptions = options.getNcaCalculationOptions();
 
     ncaOptions.setAdmsVersion("5.0.0.1");
+    ncaOptions.setProjectCategory(ProjectCategory.AGRICULTURE);
```
BertScholten, line 202: Would expect this to be validated in the asserts.

BertScholten (changes requested):

PR #305, opened 2024-10-11
Preparing for farm lodging removal

BertScholten:
Changed some default methods to non-defaults (should be implemented by now), and made some methods default and indicated that they are deprecated through annotation and comment.

Should make it easier to remove these methods in upstream project(s).


Hilbrand (approved):
LGTM

PR #306, opened 2024-10-31
AER-3268 development pressure sources

BertScholten:
Also a small cleanup/deprecation in preparation of future work.

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/NCACalculationOptions.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/NCACalculationOptions.java
@@ -17,8 +17,6 @@
 package nl.overheid.aerius.shared.domain.calculation;
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.List;
 
 /**
  * Calculation options related to the NCA theme.
@@ -27,22 +25,12 @@ public class NCACalculationOptions implements Serializable {
 
   private static final long serialVersionUID = 1L;
 
-  private ProjectCategory projectCategory;
+  private String projectCategory;
 
   /**
    * Regional area for which a permit calculation would apply to.
    */
   private String permitArea;
-
-  /**
-   * Location of the meteorological site location.
-   */
-  private String meteoSiteLocation;
-
-  /**
-   * List of meteo years to calculate.
-   */
-  private List<String> meteoYears = new ArrayList<>();
 
   /**
    * ADMS version to use.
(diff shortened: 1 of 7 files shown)


source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/ProjectCategory.java
```diff
@@ -20,7 +20,10 @@
 
 /**
  * The project category the calculation is for.
+ *
+ * @Deprecated To be replaced with database-backed category.
```
MichielJanssen-DAT, line 24: Will these database-backed categories include `GO_KARTING`, so the roundtrip test file in this PR works?

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/NCACalculationOptions.java
```diff
@@ -27,23 +27,13 @@ public class NCACalculationOptions implements Serializable {
...
-  /**
-   * Location of the meteorological site location.
-   */
-  private String meteoSiteLocation;
-
-  /**
-   * List of meteo years to calculate.
-   */
-  private List<String> meteoYears = new ArrayList<>();
-
```
MichielJanssen-DAT, line 46: Where will these be saved, if not here?

MichielJanssen-DAT (approved):
Couple questions, besides that it LGTM

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/NCACalculationOptions.java
Hilbrand: These are already in ADMSOptions. Looks like this is just some cleanup.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/NCACalculationOptions.java
```diff
@@ -27,23 +27,13 @@ public class NCACalculationOptions implements Serializable {
 
   private static final long serialVersionUID = 1L;
```
Hilbrand, line 28:
```suggestion
  private static final long serialVersionUID = 2L;
```

Hilbrand (approved):
LGTM. Just suggestion about increment serial id.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/calculation/NCACalculationOptions.java
BertScholten: Correct, the fields were not used (there were no getter/setters anymore) so should be save to remove.

PR #307, opened 2024-11-06
Avoid emission calculation for 0 vehicles

BertScholten:
Not much point in doing interpolation if the result will just be multiplied with 0 anyway.

Performance change, shouldn't have effect on results.


Hilbrand (approved):
LGTM

PR #308, opened 2024-12-05
AER-3433 keep track of calculated snapshot values

BertScholten:
Reasoning for putting it in CalculationSetOptions and the generic metadata section in IMAER is that its least changes. Conceptually it makes less sense as it's not a value that a user should generally alter/supply. But that's similar to OPS version that's also present in options/metadata I guess? Not too sure about both the reasoning and the term CalculatedSnapshotValues, suggestions are welcome.


Hilbrand (approved):
LGTM. Might be an idea to also include the timestamp of when that snapshot value was obtained.

BertScholten:
> LGTM. Might be an idea to also include the timestamp of when that snapshot value was obtained.

Myeah, might be, but would rather discuss that first before adding it. At least with this there's a clear spot for it to go.

PR #318, opened 2025-02-25
AER-3547 archive type metadata

BertScholten:
As part of this, IMAER and IMAER-java versions have changed to 6.0.1 and 6.0.1-1 respectively.


Hilbrand (approved):
LGTM

PR #319, opened 2025-03-12
AER-3597 - Support string key conversion for parsing/serialization

JornC:
Add

- toStringValue (for serialization)
- fromStringValue (for parsing)
- Tests to verify that the above 2 are reversible and symmetric

The above will allow this type to be used as a map key during serialization.

To facilitate that serialization:

- Added jackson databind dependency for necessary annotations
- Added `@JsonIgnore` to ignore `isNoYear` which is just a logical expression
- Added `@JsonKey` to `toStringValue`, which tells jackson to use that method for string key serialization
- Apparently `@JsonCreator` works for the inverse (despite jackson not using it for our use case), but it is nice to keep to that standard

### DIFF
--- a/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/sector/Sector.java
+++ b/source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/sector/Sector.java
@@ -35,7 +35,8 @@ public class Sector implements HasName, Serializable {
   public static final int DEFAULT_SECTOR_ID = 1800;
 
   /**
-   * Sector for case where no sector is specified yet. UNDEFINED MEANS UNDEFINED so don't change it into a defined sector without modifying the behavior of the application!
+   * Sector for case where no sector is specified yet. UNDEFINED MEANS UNDEFINED so don't change it into a defined sector without modifying the
+   * behavior of the application!
    */
   public static final Sector SECTOR_UNDEFINED = new Sector(0, null, "");
 


source/imaer-shared/src/test/java/nl/overheid/aerius/shared/domain/EmissionValueKeyTest.java
```diff
@@ -35,4 +36,23 @@ public void testEquals() {
...
     assertFalse(new EmissionValueKey(Substance.NH3).equals(new EmissionValueKey()), "Nothing equal");
   }
+
+  @Test
+  public void testValueConversion() {
+    final EmissionValueKey original = new EmissionValueKey(2024, Substance.NOX);
+    final String serialized = original.toStringValue();
+    final EmissionValueKey deserialized = EmissionValueKey.fromStringValue(serialized);
+
+    assertTrue(original.equals(deserialized), "Converted object should equal original");
```
BertScholten, line 46: Could just use `assertEquals` here?

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/EmissionValueKey.java
```diff
@@ -185,8 +194,35 @@ public void setSubstance(final Substance substance) {
...
+   * Expects input in the format "year:substance" as produced by
+   * {@link #toStringValue()}.
+   * 
+   * @param value The string representation in format "year:substance"
+   * @return A new EmissionValueKey instance
+   * @see #toStringValue()
+   */
+  public static EmissionValueKey fromStringValue(final String value) {
+    final String[] parts = value.split(":");
+    return new EmissionValueKey(Integer.parseInt(parts[0]), Substance.valueOf(parts[1]));
```
BertScholten, line 221: Won't this fail if you have an unexpected string as input like `"whoops"`?

JornC: ("me")

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/EmissionValueKey.java
```diff
@@ -185,8 +185,64 @@ public void setSubstance(final Substance substance) {
...
+    final String yearStr = parts.length > 0 ? parts[0].trim() : "";
+    final String substanceStr = parts.length > 1 ? parts[1].trim() : "";
+
+    if (yearStr.isEmpty() || substanceStr.isEmpty()) {
+      throw new IllegalArgumentException("Year and substance cannot be empty, got: '" + value + "'");
+    }
+
+    if (parts.length != 2) {
+      throw new IllegalArgumentException("Input must be in format 'year:substance', got: " + value);
+    }
```
BertScholten, line 231:
```suggestion
    final String[] parts = value.split(":");
    if (parts.length != 2) {
      throw new IllegalArgumentException("Input must be in format 'year:substance', got: " + value);
    }

    final String yearStr = parts.length > 0 ? parts[0].trim() : "";
    final String substanceStr = parts.length > 1 ? parts[1].trim() : "";
    if (yearStr.isEmpty() || substanceStr.isEmpty()) {
      throw new IllegalArgumentException("Year and substance cannot be empty, got: '" + value + "'");
    }
```

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/EmissionValueKey.java
```diff
@@ -185,8 +185,64 @@ public void setSubstance(final Substance substance) {
...
+      throw new NumberFormatException("Invalid year format: " + yearStr);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("Invalid substance: " + substanceStr);
+    }
+  }
+
   @Override
   public String toString() {
-    return "EmissionValueKey [year=" + year + ", substance=" + substance + "]";
+    return "EmissionValueKey [year=" + year + ", substance=" + substance.name() + "]";
```
BertScholten, line 246: think we normally use just `substance` in this case, using the normal `toString` of the enum?

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/sector/Sector.java
```diff
@@ -91,6 +94,7 @@ public OPSSourceCharacteristics getDefaultCharacteristics() {
   }
 
   @Override
+  @JsonIgnore
```
BertScholten, line 97: Wouldn't it be fine to just serialize it? In case we ever do get a different name/description combination...

source/imaer-shared/src/test/java/nl/overheid/aerius/shared/domain/EmissionValueKeyTest.java
```diff
@@ -16,23 +16,185 @@
...
-
-  @Test
-  public void testEquals() {
-    assertTrue(new EmissionValueKey(Substance.NOX).equals(new EmissionValueKey(Substance.NOX)), "Equals for NOx.");
-    assertTrue(new EmissionValueKey(Substance.NH3).equals(new EmissionValueKey(Substance.NH3)), "Equals for NH3.");
-    assertTrue(new EmissionValueKey(Substance.PM10).equals(new EmissionValueKey(Substance.PM10)), "Equals for PM10.");
-    assertFalse(new EmissionValueKey(Substance.NH3).equals(new EmissionValueKey(2020, Substance.NH3)), "No equal year");
-    assertFalse(new EmissionValueKey(Substance.NH3).equals(new EmissionValueKey(Substance.NOX)), "No equal substance");
-    assertFalse(new EmissionValueKey(Substance.NH3).equals(new EmissionValueKey()), "Nothing equal");
+@DisplayName("EmissionValueKey")
```
BertScholten, line 30: this is new... not too sure how it pans out, but guess it looks OK in the results?

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/sector/Sector.java
JornC: Yeah it would be fine, the parser-generator gizmo currently produces a warning when there is a getter without an associated field and I figured I'd just get rid of all the warnings. Will remove this one.

source/imaer-shared/src/test/java/nl/overheid/aerius/shared/domain/EmissionValueKeyTest.java
JornC: image

Yeah it looks quite nice

JornC: I'm not sure how it'll appear in eclipse but it should work fine, and it should be neatly nested aswell

JornC:
```
EmissionValueKey
├─ toString
│  ├─ should return debug format
│  └─ should handle no year case
├─ string conversion
│  ├─ should throw exception for missing separator
│  ├─ should throw exception for empty input
│  ├─ should throw exception for invalid year
│  ├─ should throw exception for invalid substance
│  ├─ should correctly serialize and deserialize
│  ├─ should throw exception for empty year
│  ├─ should throw exception for multiple separators
│  ├─ should handle whitespace
│  ├─ should throw exception for whitespace-only input
│  ├─ should throw exception for invalid format
│  ├─ should throw exception for empty substance
│  └─ should throw exception for null input
└─ equals
   ├─ should return true for identical substances with same year
   ├─ should return false for different years
   ├─ should return false for different substances
   ├─ should return false for null or empty key
   └─ should return true for identical substances with no year
```

This is how it is shown in some log file

PR #323, opened 2025-03-31
AER-3516 Implemented reading spread with legacy support

Hilbrand:
In new OPS version spread for points should be passed by the user. This means if no spread was given it should be getting half the emission height. In GML the spread can be present, but for point and lines it was not specified what the value meant as it was not used. Therefore we should treat the spread value in older GML's as a random value, and always derive the spread from the emission height. This change adds support for either reading the spread or deriving it from emission height (legacy). For now all versions of GML will use the legacy variant, as all versions are also available in production.

As of this moment there is not support for writing the spread to the GML for point/line sources, as it will not be possible to determine if it was legacy or set correctly since the GML version is not changed, and there are already GMLs generated in production with the legacy situation. To be able to correctly set the spread some new version of IMAER must be made.

### DIFF
--- a/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/characteristics/OPSSpreadUtil.java
+++ b/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/characteristics/OPSSpreadUtil.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright the State of the Netherlands
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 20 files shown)


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/characteristics/OPSSpreadUtil.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Utility to handle the legacy variant of handling the OPS spread value.
+ *
+ * In older versions of OPS the spread parameter in the input data for point sources was ignored and calculated within OPS as being half the
+ * emission height.
+ * With newer OPS versions (since 5.3.1) the spread value in the OPS input file is read and not calculated.
+ * This means the application should pass the desired spread value.
+ * For older IMAER versions (prior to 6) for point sources it means that value is derived from the emission height.
+ * Even if the spread value of a point source is set in the IMAER file it will be ignored.
+ * For none point sources and if spread was not set in older IMAER versions (< 6.0)the sector default will be used.
```
BertScholten, line 32:
```suggestion
 * For non-point sources, and if spread was not set in older IMAER versions (< 6.0), the sector default will be used.
```
(Not too sure about the commas, I just think it makes it makes sense this way...)

BertScholten:
Ah did wonder if none of our testcases would be affected, but seems like they do.

Hilbrand:
@BertScholten good point about test cases. I've updated the test cases to contain the spread value that it would use, instead of the value that was in file.

BertScholten (approved):
LGTM

PR #324, opened 2025-04-07
AER-3679 Explicitly state that maximum range is relevant

BertScholten:
As it was, the calculation method was in charge of determining if the maximum range was relevant. UK however allows users to input the maximum range for FORMAL_ASSESSMENT, so this no longer holds.
Should be safe to do it like this: the option should be explicitly set when calculating in calculator, and that should override any value that was erronously read. The default value (false) also makes sense for NL, where the only options at the moment are custom points and formal assessment anyway.


Hilbrand (approved):
LGTM

PR #325, opened 2025-04-07
Ignore a bunch of nonsense that needn't be in json

JornC:
All fields can (and should) be reconstructed using level and level1Surface

In the (eventual) parser code, this class is handled by a custom parser which does ^ that, and so all of the other fields can be ignored and left out of the json.

Added final to the fields because the GWT constructor is no longer necessary


BertScholten (approved):
ACK

JornC:
Put the commit with final fields and removing the gwt constructor in #326 pending removal of gwt serializer (which could be a while)

PR #326, opened 2025-04-07
Make fields final and remove gwt constructor

JornC:
Can be merged when the GWT serialization has been removed


BertScholten (approved):
LGTM

PR #327, opened 2025-04-08
AER-3666 Change precision for ADMS building values

BertScholten:
As far as I could tell, this is only used in frontend for display purposes.


Hilbrand (approved):
LGTM

PR #330, opened 2025-04-18
Refactor ReceptorGridSettings into an enum

Hilbrand:
Since the values of ReceptorGridSettings don't change they can be hard-coded. Breaking change: Changes `RecetorSettingsGrid` from a class into a enum.


BertScholten (approved):
LGTM

PR #335, opened 2025-05-19
AER-3735 default characteristics also determined by geometry type

BertScholten:
For ADMS it matters what the geometry type is. Or rather, it matters what the ADMS source type is, but the default for that can be based on geometry type. And since geometry type is fairly generic... Default methods in GMLCharacteristicsSupplier for easier conversion (that's the idea behind it at least).

### DIFF
--- a/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/mobile/v31/GML2OffRoad.java
+++ b/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/mobile/v31/GML2OffRoad.java
@@ -26,6 +26,7 @@ import nl.overheid.aerius.gml.base.IsGmlProperty;
 import nl.overheid.aerius.gml.base.characteristics.GML2SourceCharacteristics;
 import nl.overheid.aerius.gml.base.geo.GML2Geometry;
 import nl.overheid.aerius.gml.base.source.IsGmlEmission;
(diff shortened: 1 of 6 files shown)


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/mobile/v31/GML2OffRoad.java
```diff
@@ -99,7 +100,7 @@ private void convert(final T source, final IsGmlCustomOffRoadMobileSource custom
     newSource.setSectorId(sectorId);
     newSource.setLabel(constructLabel(source.getLabel(), customMobileSource.getDescription()));
     newSource.setCharacteristics(gml2SourceCharacteristics.fromGML(customMobileSource.getCharacteristics(),
-        getConversionData().determineDefaultCharacteristicsBySectorId(sectorId), null));
+        getConversionData().determineDefaultCharacteristicsBySectorId(sectorId, GeometryType.POINT), null));
```
Hilbrand, line 103: Wouldn't it be safer to use the geometry as set on line 110 below instead use `POINT`?

BertScholten: Good point, will change it

PR #342, opened 2025-07-17
AER-3576 Reimplemented handling spread for point/line sources

Hilbrand:
For point/line sources OPS did not use the spread. i.e. it was interpreted as being 0. Therefore for backward compatibility the spread for point/line sources for older versions of IMAER, and latest (6.0) created with AERIUS 2024 the spread will be set to 0 for point/line sources.

### DIFF
--- a/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/characteristics/OPSSpreadUtil.java
+++ b/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/characteristics/OPSSpreadUtil.java
@@ -23,15 +23,14 @@ import nl.overheid.aerius.shared.domain.v2.geojson.Point;
 /**
  * Utility to handle the legacy variant of handling the OPS spread value.
  *
- * In older versions of OPS the spread parameter in the input data for point sources was ignored and calculated within OPS as being half the
- * emission height.
- * With newer OPS versions (since 5.3.1) the spread value in the OPS input file is read and not calculated.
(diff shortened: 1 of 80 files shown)


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/characteristics/OPSSpreadUtil.java
```diff
@@ -42,27 +41,22 @@ public OPSSpreadUtil(final boolean legacySpread) {
...
 
-    if (legacySpread) {
-      if (geometry instanceof Point || geometry instanceof LineString) {
-        return getSourceSpread(null, emissionHeight, defaultSpread);
+    if (pointLine) {
+      if (legacySpread) {
+        return 0.0;
       } else {
-        return getLegacySpread(characteristics, defaultSpread);
+        return getSpreadOrDefault(characteristics, 0.0);
```
BertScholten, line 50: Shouldn't it be `defaultSpread` for the last argument? If it's a 6.0 version, but old enough AERIUS version, it'll be 0.0 anyway due to `legacySpread`, and if it's newer it makes more sense to use the supplied default as is done for polygons?

BertScholten: Myeah, I guess that could be the case, or someone creating GMLs outside of our products and not adjusting the spread. And in that case, the default from the sector makes just as much sense as 0 I imagine (just a bit weird that it works different if the geometry is a polygon). Let's stick to 0.

PR #353, opened 2025-09-25
AER-3873 - Make heat content validation more specific

JornC:
This change is actually somewhat debatable; the previous implementation was simpler and arguably more robust, however strictly speaking it was too broad by also validating the heat content when the heat content type is set to FORCED (in which case heat content is not used, but the outflow values are).

The implementation in this PR makes the validation assert precisely the valid/invalid conditions, although on the surface it is slightly more complex.


Hilbrand (approved):
LGTM

PR #355, opened 2025-09-29
Set version to 6.0.1.5-SNAPSHOT

Hilbrand:


fmadjdian (commented):
Shouldn't it be 6.0.1-5-SNAPSHOT?

fmadjdian (approved):
Looks solid

PR #373, opened 2026-03-20
Maven release github action

BertScholten:
In an attempt to automate releases a bit more, reckon that for our public repositories we might as well use a github action.

Could move part of this to our github-actions repo as well, but perhaps first lets check if it actually works the way we want?

Testing it might be a bit tricky. Currently there is no snapshot dependency, so it should just work, but would want to check that it does actually stop when there is such a dependency, for instance if the root pom used is on snapshot.

### DIFF
--- a/.github/workflows/maven-release.yml
+++ b/.github/workflows/maven-release.yml
@@ -21,6 +21,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           fetch-depth: 0
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
 
       - name: Set up Java & Maven
         uses: actions/setup-java@v5
--- a/.github/workflows/maven-release.yml
+++ b/.github/workflows/maven-release.yml
@@ -4,10 +4,10 @@ on:
   workflow_dispatch:
     inputs:
       release_version:
-        description: 'Release version (e.g. 1.2.0)'
+        description: 'Release version (e.g. 6.0.1-7)'
         required: true
       next_dev_version:
-        description: 'Next development version (e.g. 1.3.0-SNAPSHOT)'
+        description: 'Next development version (e.g. 6.0.1-8-SNAPSHOT)'
         required: true
 
 jobs:
--- a/.github/workflows/maven-release.yml
+++ b/.github/workflows/maven-release.yml
@@ -1,0 +1,71 @@
+name: Maven Release
+
+on:
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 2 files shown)


SerhatG:
@BertScholten Totally forgot about this.. Should we pick this up again?

BertScholten:
@SerhatG Yeah, we could, though not too sure what the way forward is. Looking at last commit message, GitHub Apps might be an option?

SerhatG:
> @SerhatG Yeah, we could, though not too sure what the way forward is. Looking at last commit message, GitHub Apps might be an option?

We could somewhat fix the security issues by using the GitHub Actions Policy functionality that is in Preview. I could set that only people in specific teams can run the workflow_dispatch events. I can do this organisation wide for all public repositories.
There is also functionality that only specific trusted actions are allowed to be executed.
Some projects go even further by only pinning specific SHA hashes (trusted versions of the actions), but that would mean updating them regularly (though that would be in our `github-actions` repo), so one place to manage them if we'd like.

We already use the Github App solution for Jenkins (for the publishing checks bit in our cicd-scripts). So we could do that as well (or a combination of both)·.

Shall I pick this up and see if I can make something useful based on your version?

BertScholten:
Only having people in specific teams be able to do it sounds good. Using a functionality in preview is a bit risky (might be a bit volatile), but it would beat how we're doing it currently I think.

Pinning specific SHA hashes is something that is done at DAT as well, though not entirely sure how that's being kept up to date. Vaguely remember something about dependabot doing checks on that, could have a look if you want?

I'd be fine with you picking this up, I don't have the permissions to change anything in settings anyway so all I can do is fiddle around on my own fork ;)

SerhatG:
> Only having people in specific teams be able to do it sounds good. Using a functionality in preview is a bit risky (might be a bit volatile), but it would beat how we're doing it currently I think.
> 
> Pinning specific SHA hashes is something that is done at DAT as well, though not entirely sure how that's being kept up to date. Vaguely remember something about dependabot doing checks on that, could have a look if you want?
> 
> I'd be fine with you picking this up, I don't have the permissions to change anything in settings anyway so all I can do is fiddle around on my own fork ;)

Yeah, I believe if the SHA pinning is done right, dependabot supports that and can come with update PRs.

I believe usually the preview functionality just gets marked as final, but yeah alternative is not doing anything there anyway. So might as well start somewhere.

BertScholten:
Superseded by #392

PR #379, opened 2026-04-02
AER-4271 Added support for using max speed for NON_URBAN ROADS

Hilbrand:
With new road emissions for NON_URBAN_ROAD speed will be relevant. Therefore speed will be read. Because with new data speed determines the emission and not only the road type code a default speed will be set if the speed is null or 0. The speed set is representative for the road type. For  NON_URBAN_ROAD_NATIONAL  it will be 80, for NON_URBAN_ROAD_GENERAL it will be 60. Calculator 2025 doesn't use these speed values yet, but having set them has no impact, but it also isn't effecting anything.

Simplified GML2SRMRoad v4 implementation by merging specific variables with convertRoadType because specificVariables was only used for road type, and since both methods have a SRM specific implementation the SRM1 case can be moved to the convertRoadTypeCode.


willvoo (approved):
LGTM

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/road/v40/GML2SRMRoad.java
```diff
@@ -112,6 +118,7 @@ protected void addEmissionValues(final List<Vehicles> addToVehicles, final T sou
     standardVehicle.getValuesPerVehicleTypes().put(sv.getVehicleType().getStandardVehicleCode(), valuesPerVehicleType);
   }
 
+
```
BertScholten:
```suggestion
```

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/road/v40/GML2SRMRoad.java
```diff
@@ -57,15 +57,16 @@ protected GML2SRMRoad(final GMLConversionData conversionData) {
...
+    emissionSource.setRoadTypeCode(convertRoadTypeCode(roadType));
     for (final IsGmlProperty<IsGmlVehicle> vp : source.getVehicles()) {
-      addVehicleEmissions(emissionSource.getSubSources(), source, vp, mergingStandardVehicles);
+      addVehicleEmissions(roadType, emissionSource.getSubSources(), source, vp, mergingStandardVehicles);
     }
     emissionSource.setTrafficDirection(source.getTrafficDirection());
     emissionSource.setRoadManager(source.getRoadManager());
     emissionSource.setRoadAreaCode("NL");
     // Ensure road type get set before specific, as it's overwritten by SRM1
```
BertScholten: Might make more sense to move this comment as well, or remove it completely.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v6_0/togml/Road2GML.java
```diff
@@ -137,42 +139,54 @@ private List<VehiclesProperty> toVehicleProperties(final List<Vehicles> vehicleG
...
-  private void handleTunnel(final nl.overheid.aerius.shared.domain.v2.source.RoadEmissionSource emissionSource, final SRM2Road returnSource) {
+  private static String gmlRoadTypeCode(final String roadTypeCode, final List<Vehicles> vehicles) {
+    // If there are standard vehicles with no speed set keep the original road type code for non urban roads.
+    // This to be backward compatible for when the code is used in combination with data not yet supporting non urban roads with max speed.
+    // Because if NON_URBAN_ROAD would be used in combination with no speed, it would mean it would be interpreted as having 0 speed,
+    // and it could mean using an invalid speed category because it would have been expected the speed to be 80 km/h.
+    final boolean maxSpeedNotSet =  vehicles.stream().filter(StandardVehicles.class::isInstance).map(StandardVehicles.class::cast)
+        .anyMatch(e -> Optional.ofNullable(e.getMaximumSpeed()).orElse(0) == 0);
+
+    return !maxSpeedNotSet && roadTypeCode.startsWith(RoadType.NON_URBAN_ROAD.name()) ? RoadType.NON_URBAN_ROAD.name() : roadTypeCode;
```
BertScholten: This means that if we have a release of calculator before next actualisation release (when all these sources should get a max speed), all exported GMLs for non urban roads will get `NON_URBAN_ROAD` as road type? And after the next actualisation release they once again will use `NON_URBAN_ROAD_GENERAL` or `NON_URBAN_ROAD_NATIONAL`?

Hope no other release in that case.

BertScholten (approved):
Not a fan of the temporary bit, but otherwise LGTM

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/road/v40/GML2SRMRoad.java
Hilbrand: A good point. This can actually be simplified by just moving the code that is set in the specific in the convertRoadTypeCode

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v6_0/togml/Road2GML.java
Hilbrand: No (at least that is not the idea of this implementation). At any moment (and this can be before the actualisation release when the current data would be used) if no speed is set or if speed is 0 it will keep the `NON_URBAN_ROAD_GENERAL` or `NON_URBAN_ROAD_NATIONAL` (those are used internally to specify road type). At any point when it would have a speed > 0 it will set `NON_URBAN_ROAD`. In the future unless we get non urban road with no speed again the `NON_URBAN_ROAD_GENERAL` or `NON_URBAN_ROAD_NATIONAL` should not be set anymore.
The idea is in GML it can have:
- no speed or 0 -> `NON_URBAN_ROAD_GENERAL` or `NON_URBAN_ROAD_NATIONAL`
- with speed > 0 -> `NON_URBAN_ROAD`

Does this seem to match with what you see in the code?

I'll rewrite the boolean check to be inverted: instead of `any` use `all`. That will make it a bit more readable.

BertScholten: Ah OK, the doube negation did get me there. But if that's the case, then the codes in the database no longer match those in GML, that seems a bit of a drawback? Or did that get changed in later data supply? (ones I got so far kept the same road types).

And when importing its setting a max speed anyway, so there is already going something awry anyway if you export->import->export? If you create a new source in calculator now, it'll get `NON_URBAN_ROAD_NATIONAL` with 0 max speed in the exported GML. If you then import it gets max speed 80 due to `GML2Road`, and when you export that again you get `NON_URBAN_ROAD` with 80 max speed?

Hilbrand: I added the `NON_URBAN_ROAD` to make explicit separation between old and new. old is without speed and new with speed. That way when `NON_URBAN_ROAD` is present it indicates a speed is to be set. But on second thought this might not add much. Therefore I removed `NON_URBAN_ROAD` and kept the original only. This still means when reading old files with speed not set or 0 it will get a speed value (60 or 80) because when new emission factors are present it must have a speed. To be already compatible with that situation it will already set the speed when importing. For current behaviour this has no impact, other than when exporting the new speed value is set. Newly created sources in the UI do get 0 currently and 0 will be writtten. But that is no problem. Because when imported again (in the future) the speed will be set.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/road/GML2Road.java
```diff
@@ -48,13 +48,15 @@ protected GML2Road(final GMLConversionData conversionData) {
   public S convert(final T source) throws AeriusException {
     final S emissionSource = construct();
     final List<StandardVehicles> mergingStandardVehicles = new ArrayList<>();
+    final String roadTypeCode = source.getRoadTypeCode();
+
     for (final IsGmlProperty<IsGmlVehicle> vp : source.getVehicles()) {
-      addVehicleEmissions(emissionSource.getSubSources(), source, vp, mergingStandardVehicles);
+      addVehicleEmissions(source.getRoadTypeCode(), emissionSource.getSubSources(), source, vp, mergingStandardVehicles);
```
BertScholten:
```suggestion
      addVehicleEmissions(roadTypeCode, emissionSource.getSubSources(), source, vp, mergingStandardVehicles);
```

BertScholten (approved):
LGTM

PR #381, opened 2026-05-04
AER-3162 - Group XSD validation errors by source location

JornC:
Validation events from JAXB are now grouped by their (line, column) locator so a single bad value no longer produces multiple separate errors.

Within each group only the highest-severity events are kept and their messages combined. Each combined message is prefixed with [line N, col M] so users can locate the offending value in larger GML files.

### DIFF
--- a/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLReaderFactory.java
+++ b/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLReaderFactory.java
@@ -172,10 +172,6 @@ public final class GMLReaderFactory {
   }
 
   private static List<AeriusException> newValidationFailed(final List<ValidationEvent> list) {
-    // Group events that share a source location (line + column), discard the ones whose
-    // information is already covered by a higher-severity event at the same location, then
-    // combine the survivors into one message per location. Each combined message is prefixed
-    // with [line N, col M] so users can find the offending value.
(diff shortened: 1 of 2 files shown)


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/GMLReaderFactory.java
```diff
@@ -167,13 +172,54 @@ private static AeriusException newGmlParseError(final XMLStreamException e) {
...
+      grouped.computeIfAbsent(locationKey(event.getLocator()), k -> new ArrayList<>()).add(event);
+    }
+
+    final List<AeriusException> errors = new ArrayList<>();
+    for (final List<ValidationEvent> group : grouped.values()) {
+      final List<ValidationEvent> kept = discardRedundantLowerSeverityEvents(group);
+      final String body = kept.stream()
+          .map(e -> e.getMessage().trim())
+          .collect(Collectors.joining(" "));
+      final String message = locationPrefix(kept.get(0).getLocator()) + body;
```
Hilbrand: Can `kept` also be empty here?

JornC: No, these are groups with a minimal list size of 1.

However it's worth it to refactor this to avoid having to come to that conclusion

Hilbrand (approved):
LGTM

PR #382, opened 2026-06-04
AER-4271 When importing Non Urban Road with multiple vehicles to check if can be combined check on converted maximumspeed

Hilbrand:
Because when creating a StandardVehicles source it changes the maximum speed for the source for non urban roads. Than the other vehicle sources should be compared to that speed to check on matching, otherwise it will try to match a non converted with converted source which will result in the vehicle source to be separated. Updated round trip sources to include 2 different vehicle type sources. Note that the round trip test did catch this case because when exporting it will merge again. Therefore the export was correct, and wasn't detecting the issue.


JornC (approved):

BertScholten (approved):
LGTM

PR #383, opened 2026-06-16
AER-3979 Moved classes that are not used in UI from shared to util module

Hilbrand:
This because classes in shared are intended for usage also in UI and to keep code separated. Kept them in the same package to keep it backward compatible.


BertScholten (approved):
ACK

PR #384, opened 2026-06-16
AER-3979 Refactored Offroad mobile to reduce duplications

Hilbrand:


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/mobile/AbstractGML2OffRoad.java
```diff
@@ -0,0 +1,93 @@
...
+import nl.overheid.aerius.shared.exception.ImaerExceptionReason;
+
+/**
+ * Base class for converting OffRoad mobile from GML to internal data object.
+ *
+ * @param <S> sub class of the standard offroad mobile sources
+ * @param <T> sub class of the off road mobile source
+ */
+public abstract class AbstractGML2OffRoad<S extends IsGmlStandardOffRoadMobileBaseSource, T extends IsGmlOffRoadMobileEmissionSource>
+extends AbstractGML2Specific<T, OffRoadMobileEmissionSource> {
```
BertScholten:
```suggestion
public abstract class AbstractGML2OffRoad<S extends IsGmlStandardOffRoadMobileBaseSource, T extends IsGmlOffRoadMobileEmissionSource>
    extends AbstractGML2Specific<T, OffRoadMobileEmissionSource> {
```

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/mobile/AbstractGML2OffRoad.java
```diff
@@ -0,0 +1,93 @@
...
+import nl.overheid.aerius.shared.exception.AeriusException;
+import nl.overheid.aerius.shared.exception.ImaerExceptionReason;
+
+/**
+ * Base class for converting OffRoad mobile from GML to internal data object.
+ *
+ * @param <S> sub class of the standard offroad mobile sources
+ * @param <T> sub class of the off road mobile source
+ */
+public abstract class AbstractGML2OffRoad<S extends IsGmlStandardOffRoadMobileBaseSource, T extends IsGmlOffRoadMobileEmissionSource>
```
BertScholten: With `IsGmlOffRoadMobileEmissionSource` being the main source, and `IsGmlStandardOffRoadMobileBaseSource` being one of the subsource implementations, swapping the 2 around might be a bit more logical? Bit nitpicky point though, fine with this as well.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v1_0/source/mobile/OffRoadMobileSource.java
```diff
@@ -32,16 +32,16 @@
 @XmlType(name = "StandardOffRoadMobileSourceType", namespace = CalculatorSchema.NAMESPACE)
 public class OffRoadMobileSource extends AbstractOffRoadMobileSource implements IsGmlStandardOffRoadMobileSource {
 
-  private int literFuelPerYear;
+  private Integer literFuelPerYear;
```
BertScholten, line 35: Imagine the XSD prevents this actually being `null`, and we're not writing these old values anymore, so in practice it'll still be `int`? And it's nice to get rid of some duplications.

BertScholten (approved):
Some code style suggestions, but otherwise LGTM

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/mobile/AbstractGML2OffRoad.java
Hilbrand: Yes `IsGmlOffRoadMobileEmissionSource` is the main source, and the sub class `IsGmlStandardOffRoadMobileBaseSource`. But this specific generic is for sub classes of the `Standard` therefor use the sub class `IsGmlStandardOffRoadMobileBaseSource`. Also since the custom is also a sub class of `IsGmlOffRoadMobileEmissionSource`. Which would make it a bit to generic to use `IsGmlOffRoadMobileEmissionSource` here.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/v1_0/source/mobile/OffRoadMobileSource.java
Hilbrand: Yes for reading it does not really matter, and it makes it possible to have a shared interface.

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/mobile/AbstractGML2OffRoad.java
BertScholten: My intention was to switch the 2 types around, so `AbstractGML2OffRoad<T extends IsGmlOffRoadMobileEmissionSource, S extends IsGmlStandardOffRoadMobileBaseSource>`, just so it reads `<main source class, specific sub source class>`).

Perhaps renaming `IsGmlOffRoadMobileSource` to `IsGmlOffRoadMobileSubSource` would also be an option, but that would make it a bit more inconsistent with the rest 🙈

Hilbrand: Ah I see what you mean. Good idea. I changed the order. The rename to include 'Sub' would make it inconsistent with the rest indeed, and than probably a lot more should be renamed. So I'll keep it as is for now.

PR #386, opened 2026-06-23
AER-4514 Change max roadspeed to default to 60 for non urban roads

Hilbrand:
As speed has become relevant as there will be different emission factors for speed, a default speed needs to be set for older GML's this was earlier set to 80, but should be 60. Also added a warning code in case this default value is set.

### DIFF
--- a/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/road/v10/GML2SRM2RoadV10.java
+++ b/source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/road/v10/GML2SRM2RoadV10.java
@@ -17,6 +17,7 @@
 package nl.overheid.aerius.gml.base.source.road.v10;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 
 import nl.overheid.aerius.gml.base.GMLConversionData;
@@ -45,10 +46,10 @@ public class GML2SRM2RoadV10<T extends IsGmlSRM2RoadOld> extends GML2SRM2RoadV11
   @Override
   protected void addEmissionValues(final RoadType roadType, final List<Vehicles> addToVehicles, final T source, final IsGmlStandardVehicle sv,
       final List<StandardVehicles> mergingStandardVehicles) {
-    final StandardVehicles standardVehicle = findExistingMatch(sv, mergingStandardVehicles).orElseGet(() -> {
+    final StandardVehicles standardVehicle = findExistingMatch(sv, roadType, mergingStandardVehicles).orElseGet(() -> {
       final StandardVehicles vse = new StandardVehicles();
 
-      vse.setMaximumSpeed(getMaximumSpeed(roadType, source.getMaximumSpeed()));
+      vse.setMaximumSpeed(getMaximumSpeed(source.getId(), true, roadType, source.getMaximumSpeed()));
       vse.setStrictEnforcement(source.isStrictEnforcement());
       vse.setTimeUnit(TimeUnit.valueOf(sv.getTimeUnit().name()));
       mergingStandardVehicles.add(vse);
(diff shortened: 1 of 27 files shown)


source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/road/v10/GML2SRM2RoadV10.java
```diff
@@ -62,11 +63,13 @@ protected void addEmissionValues(final RoadType roadType, final List<Vehicles> a
   }
 
   @Override
-  protected Optional<StandardVehicles> findExistingMatch(final IsGmlStandardVehicle sv, final List<StandardVehicles> mergingStandardVehicles) {
+  protected Optional<StandardVehicles> findExistingMatch(final IsGmlStandardVehicle sv, final RoadType roadType,
+      final List<StandardVehicles> mergingStandardVehicles) {
     return mergingStandardVehicles.stream()
+        .filter(x -> Objects.equals(x.getMaximumSpeed(), getMaximumSpeed(null, false, roadType, sv.getMaximumSpeed())))
+        .filter(x -> Boolean.TRUE.equals(x.getStrictEnforcement()) == Boolean.TRUE.equals((sv.isStrictEnforcement())))
```
BertScholten: I guess this works, but for v1.0 it isn't exactly required as at that point the max speed and strict enforcement were defined on the source level instead of per sub source. So these predicates will always end up being true (see ` nl.overheid.aerius.gml.v1_0.source.road.StandardVehicle`).

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/base/source/road/GML2RoadTest.java
```diff
@@ -17,71 +17,218 @@
...
-  /**
-   * Test merging sub sources. vehicle types with the same configuration can be joined into a single sub source.
-   */
+  @BeforeEach
+  void beforeEach() {
+    warnings = new ArrayList<>();
+    lenient().doReturn(warnings).when(conversionData).getWarnings();
+  }
+
+  // Test for GML version 1.0 and below.
```
BertScholten, line 72: Not sure if it's worth it, but could make a `GML2SRM2RoadV10Test` class instead? Would make it easier to clean this out as well whenever we get around to cleaning up old GML versions...

source/imaer-gml/src/test/resources/gml/v0_5/roundtrip/road_dynamic_segmentation.gml
```diff
@@ -54,18 +54,18 @@
...
                     <imaer:vehiclesPerDay>200.0</imaer:vehiclesPerDay>
                     <imaer:stagnationFactor>0.0</imaer:stagnationFactor>
                 </imaer:StandardVehicle>
             </imaer:vehicles>
+            <imaer:vehicles>
+                <imaer:StandardVehicle vehicleType="LIGHT_TRAFFIC">
+                    <imaer:vehiclesPerDay>980.0</imaer:vehiclesPerDay>
+                    <imaer:stagnationFactor>0.2</imaer:stagnationFactor>
+                </imaer:StandardVehicle>
+            </imaer:vehicles>
```
BertScholten, line 62: Not entirely sure why this change is required? Wouldn't expect ordering in these old versions to have an effect on the newer versions at least. Or does that new filtering thing cause them to not end up in the same group anymore?

BertScholten (changes requested):

source/imaer-gml/src/main/java/nl/overheid/aerius/gml/base/source/road/v10/GML2SRM2RoadV10.java
Hilbrand: I had a second look. I added to be the same as others. But strict and max are not present so in sub sources so should always return true. Same is also for timeunit. Therefore I removed the checks. The max speed check even caused it to be split up into to much sub sources. So I've also updated the unit test.

source/imaer-gml/src/test/java/nl/overheid/aerius/gml/base/source/road/GML2RoadTest.java
Hilbrand: I thought about it too. But because I have some generic code it would either mean to duplicate it or add an additional shared super class. I thought this would be a bit too much additional code. Therefore I kept it in a single class. I had already added comment blocks to separate the different versions. But will also move the version specific create methods to those blocks instead of being in the generic part below to have is separated better.

source/imaer-gml/src/test/resources/gml/v0_5/roundtrip/road_dynamic_segmentation.gml
Hilbrand: This was indeed due to filtering. But the filtering was not correct. With this being fixed this change will be reverted.

BertScholten: File (and 3 other v0_5 and v1_0 files) are still changed in the PR, guess the order in these files doesn't actually matter for the tests.

BertScholten (approved):
LGTM

source/imaer-gml/src/test/resources/gml/v0_5/roundtrip/road_dynamic_segmentation.gml
Hilbrand: Yes you're right. I've reverted those changes as they are not needed.

PR #389, opened 2026-07-16
AER-4540 Added emission calculation and source validation related to adding power support (U-method) to off road mobile sources.

Hilbrand:
This change doesn't add support in IMAER. Meaning if power value is present it's not yet persisted in the GML or read from a GML.


source/imaer-gml/src/test/java/nl/overheid/aerius/test/TestValidationAndEmissionHelper.java
```diff
@@ -149,21 +149,20 @@ public class TestValidationAndEmissionHelper implements ValidationHelper, Emissi
...
+          null),
+      new OffRoadConstructHelper("SIIIA5675DSN", new EmissionHelper(0.025, 0.00024), new EmissionHelper(0.005, 0.0), null, null),
+      new OffRoadConstructHelper("SIIIB5675DSN", new EmissionHelper(0.027, 0.00024), new EmissionHelper(0.005, 0.0), null, null),
+      new OffRoadConstructHelper("SIIIA75560DSN", new EmissionHelper(0.002, 0.00024), new EmissionHelper(0.005, 0.0), null, null),
+      new OffRoadConstructHelper("SII5675DSN", new EmissionHelper(0.005, 0.00024), new EmissionHelper(0.005, 0.0), null, null),
+      new OffRoadConstructHelper("SI5675DSN", new EmissionHelper(0.035, 0.00024), new EmissionHelper(0.005, 0.0), null, null),
+      new OffRoadConstructHelper("SI56DSN", new EmissionHelper(0.045, 0.00024), new EmissionHelper(0.005, 0.0), null, null),
+      new OffRoadConstructHelper("SIV75560DSJ", new EmissionHelper(0.055, 0.00024), new EmissionHelper(0.005, 0.0), null, null),
+      new OffRoadConstructHelper("SIIIA56DSN", new EmissionHelper(0.040, 0.00024), new EmissionHelper(0.005, 0.0), null, null));
```
BertScholten, line 165: Don't need a category that does have emission factors for power?

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/offroad/StandardOffRoadMobileSource.java
```diff
@@ -16,14 +16,29 @@
...
+ * The most common emission calculation methods are related to Stage mobile equipment.
+ * For stage classes there are 2 methods:
+ *
+ * <ol>
+ * <li>Adblue(adblue) Uren(operation hours) Brandstof(fuel) or AUB-method.
+ * <li>Uren(operation hours) or U-method, which is based on machine power(kW).
+ * </ol>
+ *
+ * The convention for calculating emissions for this object is as follows for off road categories that support both power and fuel bases emissions.
+ * If the power is specified the power value should be used. Otherwise the fuel value will be used.
```
BertScholten, line 30: Don't want to use an explicit property to make the distinction? As the user also selects which method to use, would have expected that to be the case.

source/imaer-util/src/main/java/nl/overheid/aerius/shared/emissions/OffRoadMobileEmissionFactorSupplier.java
```diff
@@ -43,4 +43,10 @@ public interface OffRoadMobileEmissionFactorSupplier {
    */
   Map<Substance, Double> getOffRoadMobileEmissionFactorsPerLiterAdBlue(final String offRoadMobileSourceCode);
 
+  /**
+   * Obtain emission factors per engine kiloWatt/hour.
+   */
+  default Map<Substance, Double> getOffRoadMobileEmissionFactorsPerKW(final String offRoadMobileSourceCode) {
```
BertScholten: `PerKWh`? or `PerKiloWattHour`? Or even `PerKiloWattPerOperatingHour`, as it is a combination of engine power and operating hours in this case, but that does end up with quite a long method name...

source/imaer-util/src/main/java/nl/overheid/aerius/shared/emissions/OffRoadMobileEmissionsCalculator.java
```diff
@@ -75,38 +78,56 @@ Map<Substance, BigDecimal> calculateEmissions(final CustomOffRoadMobileSource cu
...
+      final long operatingHours) {
+    final Map<Substance, BigDecimal> emissions = new EnumMap<>(Substance.class);
 
-    // AdBlue factors are (/can be) negative. To ensure we don't end up with negative emissions, ensure all negative values are set to 0.
-    // This should have been checked beforehand, so more of a precaution.
-    for (final Substance substance : Substance.values()) {
-      emissions.computeIfPresent(substance, (subst, value) -> value.compareTo(BigDecimal.ZERO) < 0 ? BigDecimal.ZERO : value);
-    }
+    kWPerYear.ifPresent(power -> {
+      final BigDecimal operatingHoursDivided = BigDecimal.valueOf(operatingHours).movePointLeft(DIVIDE_BY_1000);
```
BertScholten, line 114: I don't see why operating hours get divided. Isn't the the formula just `emission in kg/y = operating hours in h/y * power in kW * emission factor in kg/kW/h`? or are emission factors in g/kW/h`?

source/imaer-util/src/main/java/nl/overheid/aerius/shared/emissions/OffRoadMobileEmissionsCalculator.java
```diff
@@ -75,38 +78,56 @@ Map<Substance, BigDecimal> calculateEmissions(final CustomOffRoadMobileSource cu
...
+      final Map<Substance, Double> emissionFactorsPerKW = emissionFactorSupplier.getOffRoadMobileEmissionFactorsPerKW(mobileSourceCode);
 
+      emissionFactorsPerKW.forEach((key, value) ->
+      emissions.put(key, BigDecimal.valueOf(value).multiply(power).multiply(operatingHoursDivided)));
```
BertScholten: indentation seems off

source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidationHelper.java
```diff
@@ -18,16 +18,80 @@
...
+  }
+
+  /**
+   * Returns true if the given power value is within the off road category power range.
+   *
+   * @param offRoadMobileSourceCode code of the off road source
+   * @param powerValue power value to check
+   * @return true if within range
+   */
+  default boolean isPowerWithinRange(final String offRoadMobileSourceCode, final int powerValue) {
```
BertScholten: mmm might make more sense to retrieve the range to implement the check in the validator rather than relying on the caller to do this properly? Would also prevent the need for the next method I think?

source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
```diff
@@ -62,24 +62,45 @@ private boolean validateStandardOffRoad(final StandardOffRoadMobileSource subSou
...
+  private boolean validatePowerOrLiterFuel(final StandardOffRoadMobileSource subSource) {
     boolean valid = true;
-    if (validationHelper.expectsLiterFuelPerYear(subSource.getOffRoadMobileSourceCode()) && subSource.getLiterFuelPerYear() == null) {
-      getErrors().add(new AeriusException(ImaerExceptionReason.MOBILE_SOURCE_MISSING_LITER_FUEL, subSource.getDescription()));
+    final String code = subSource.getOffRoadMobileSourceCode();
+    final boolean expectsPower = validationHelper.expectsPower(code);
+    final boolean noPower = expectsPower && subSource.getPower() == null;
+    final boolean expectsFuel = validationHelper.expectsLiterFuelPerYear(code);
+    final boolean noFuel = expectsFuel && subSource.getLiterFuelPerYear() == null;
+    if (noPower && noFuel) {
```
BertScholten: Don't think this works for the 'utiliteitsvoertuigen' which only expect operating hours as input?
(and yeah, not too sure that case is covered by tests)

source/imaer-util/src/test/java/nl/overheid/aerius/validation/OffRoadValidatorTest.java
```diff
@@ -294,19 +289,20 @@ private OffRoadMobileEmissionSource constructSource() {
...
 
   private void mockCategoryExpectingAll(final String code) {
-    mockCategoryExpectingLiterFuel(code);
+    mockCategoryExpectingPowerAndLiterFuel(code);
     mockCategoryExpectingOperatingHours(code);
     mockCategoryExpectingLiterAdBlue(code);
   }
 
-  private void mockCategoryExpectingLiterFuel(final String code) {
+  private void mockCategoryExpectingPowerAndLiterFuel(final String code) {
```
BertScholten: Think there will also be the case where only liter fuel is expected? And for completeness sake it might be good to have a test case where the category only expects power?

BertScholten (changes requested):

source/imaer-gml/src/test/java/nl/overheid/aerius/test/TestValidationAndEmissionHelper.java
Hilbrand: Yes likely, but these values itself are not yet used. They'll be used in the GML tests. But since IMAER is not yet adapted it's not useful to put in a random value.

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/offroad/StandardOffRoadMobileSource.java
Hilbrand: No. The idea was to not store the method selected. So it's not needed to make it explicit here.

source/imaer-util/src/main/java/nl/overheid/aerius/shared/emissions/OffRoadMobileEmissionsCalculator.java
Hilbrand: Yes. If the information in the item is correct the emission factor is in `g/kW/h`.

source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
Hilbrand: The variable names might be a bit misleading here. Because `noPower` means the category has power emission factors, but no power value specified. If the same is true for `noFuel` it means no value was set for fuel. This case works by assuming when power emission factors are available fuel is also available. But that might theoretically not the case. So I'll have a second look into improving this.

source/imaer-util/src/main/java/nl/overheid/aerius/shared/emissions/OffRoadMobileEmissionFactorSupplier.java
```diff
@@ -43,4 +43,10 @@ public interface OffRoadMobileEmissionFactorSupplier {
    */
   Map<Substance, Double> getOffRoadMobileEmissionFactorsPerLiterAdBlue(final String offRoadMobileSourceCode);
 
+  /**
+   * Obtain emission factors per engine kiloWatt/hour.
```
BertScholten:
```suggestion
   * Obtain emission factors per engine kiloWatt and per operating hour (in gram/kilowatt/hour).
```
Specifying the unit here is quite important with it being in grams where the others are in kilograms. (not sure if that's something that can be discussed with the actualisation team, how the units are not really consistent?)

source/imaer-shared/src/main/java/nl/overheid/aerius/shared/domain/v2/source/offroad/StandardOffRoadMobileSource.java
BertScholten: mmk, think I'd have made it explicit so the validation is also not based on assumptions, but I guess it can work like this.

BertScholten (approved):
LGTM, just a note about units which is a bit inconsistent but is also slightly out of our control (unless you do the conversion from gram to kilogram on the calculator side).

PR #390, opened 2026-07-20
AER-4529 Power attribute in IMAER for mobile sources

BertScholten:
Could have done something complex with different types of implementations of `StandardOffRoadMobileSource`, but that wouldn't have made this an incremental update, so went with the same approach as is used in the code. Let validation figure it out.

As part of this, also switching IMAER-java to version `6.0.2-1-SNAPSHOT`, as IMAER switches to `6.0.2`.

(Turns out we didn't update the incremental version in the actual XSD to 6.0.1 🙈 )


Hilbrand (approved):
LGTM

PR #391, opened 2026-07-21
[github-actions] Use new init function

SerhatG:
Gets rid of the custom checkout that we need to keep updated and optionally multiple references to the version of (aerius-)github-actions.


BertScholten (approved):
ACK

PR #395, opened 2026-08-03
AER-4568 Fix import

MichielJanssen-DAT:
- Fix import by not requiring adblue for power based method
- Clean up the OffRoadValidator a bit
- Unit test updates

### DIFF
--- a/source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
+++ b/source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
@@ -135,6 +135,7 @@ class OffRoadValidator extends SourceValidator<OffRoadMobileEmissionSource> {
   }
 
   private void validateOffRoadLiterAdBlue(final StandardOffRoadMobileSource subSource) {
+    // No need for null check, as validateAUBMethod() handles that if required
     if (subSource.getLiterAdBluePerYear() != null && subSource.getLiterAdBluePerYear() == 0) {
       subSource.setLiterAdBluePerYear(null);
     } else if (subSource.getLiterAdBluePerYear() != null && subSource.getLiterFuelPerYear() != null
--- a/source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
+++ b/source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
@@ -116,14 +116,10 @@ class OffRoadValidator extends SourceValidator<OffRoadMobileEmissionSource> {
     if (!expectsFuel) {
       subSource.setLiterFuelPerYear(null);
     }
-    if (validationHelper.expectsLiterAdBluePerYear(code) && subSource.getLiterAdBluePerYear() == null) {
-      getErrors().add(new AeriusException(ImaerExceptionReason.MOBILE_SOURCE_MISSING_LITER_ADBLUE, subSource.getDescription()));
-      valid = false;
-    }
     // Fallback if custom GML is passed with power = 0
     subSource.setPower(null);
 
-    validateOffRoadLiterAdBlue(subSource);
+    valid = validateOffRoadLiterAdBlue(subSource) && valid;
     return valid;
   }
 
@@ -153,14 +149,19 @@ class OffRoadValidator extends SourceValidator<OffRoadMobileEmissionSource> {
     return valid;
   }
 
-  private void validateOffRoadLiterAdBlue(final StandardOffRoadMobileSource subSource) {
-    // No need for null check, as validateFuelBased() handles that if required
-    if (subSource.getLiterAdBluePerYear() != null && subSource.getLiterAdBluePerYear() == 0) {
+private boolean validateOffRoadLiterAdBlue(final StandardOffRoadMobileSource subSource) {
+    boolean valid = true;
+    if (validationHelper.expectsLiterAdBluePerYear(subSource.getOffRoadMobileSourceCode())) {
+      if (subSource.getLiterAdBluePerYear() == null) {
+        getErrors().add(new AeriusException(ImaerExceptionReason.MOBILE_SOURCE_MISSING_LITER_ADBLUE, subSource.getDescription()));
+        valid = false;
+      } else if (subSource.getLiterFuelPerYear() != null) {
+        validateAdBlueFuelRatio(subSource);
+      }
+    } else if (subSource.getLiterAdBluePerYear() != null && subSource.getLiterAdBluePerYear() == 0) {
       subSource.setLiterAdBluePerYear(null);
-    } else if (subSource.getLiterAdBluePerYear() != null && subSource.getLiterFuelPerYear() != null
-        && validationHelper.expectsLiterAdBluePerYear(subSource.getOffRoadMobileSourceCode())) {
-      validateAdBlueFuelRatio(subSource);
     }
+    return valid;
   }
 
   private void validateAdBlueFuelRatio(final StandardOffRoadMobileSource subSource) {
--- a/source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
+++ b/source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
@@ -83,11 +83,15 @@ class OffRoadValidator extends SourceValidator<OffRoadMobileEmissionSource> {
   private EmissionMethod determineEmissionMethod(final StandardOffRoadMobileSource subSource) {
     final String code = subSource.getOffRoadMobileSourceCode();
     final boolean hasPower = subSource.getPower() != null && subSource.getPower() > 0;
+    final boolean hasFuel = subSource.getLiterFuelPerYear() != null && subSource.getLiterFuelPerYear() > 0;
     // A category can accept power or fuel. At least one is mandatory.
-    if (hasPower) {
-      return validationHelper.expectsPower(code) ? EmissionMethod.POWER : EmissionMethod.MISSING;
+    if (hasPower && validationHelper.expectsPower(code)) {
+      return EmissionMethod.POWER;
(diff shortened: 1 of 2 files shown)


source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
```diff
@@ -65,30 +65,46 @@ private boolean validateStandardOffRoad(final StandardOffRoadMobileSource subSou
...
+    validateOffRoadLiterAdBlue(subSource);
+    return valid;
   }
 
-  private boolean validatePowerOrLiterFuel(final StandardOffRoadMobileSource subSource) {
+  private boolean validatePowerBased(final StandardOffRoadMobileSource subSource) {
+    boolean valid = true;
+    final String code = subSource.getOffRoadMobileSourceCode();
+    if (!validationHelper.expectsPower(code)) {
+      getErrors().add(new AeriusException(ImaerExceptionReason.MOBILE_SOURCE_MISSING_POWER_OR_LITER_FUEL, subSource.getDescription()));
```
BertScholten: This is now a bit weird exception, as the user has supplied power, but apparantly that is not something used with this category. Should it error out in this case, or should it error out when either is expected, but neither are supplied?

source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
```diff
@@ -118,19 +134,14 @@ private boolean validateOffRoadOperatingHours(final StandardOffRoadMobileSource
...
-      }
-    } else if (subSource.getLiterAdBluePerYear() != null && subSource.getLiterAdBluePerYear() == 0) {
+  private void validateOffRoadLiterAdBlue(final StandardOffRoadMobileSource subSource) {
+    // No need for null check, as validateFuelBased() handles that if required
+    if (subSource.getLiterAdBluePerYear() != null && subSource.getLiterAdBluePerYear() == 0) {
       subSource.setLiterAdBluePerYear(null);
+    } else if (subSource.getLiterAdBluePerYear() != null && subSource.getLiterFuelPerYear() != null
+        && validationHelper.expectsLiterAdBluePerYear(subSource.getOffRoadMobileSourceCode())) {
+      validateAdBlueFuelRatio(subSource);
     }
```
BertScholten: Might make more sense to call this validation from `validateFuelBased`? For power-based method it serves no purpose at least.

BertScholten (changes requested):

source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
MichielJanssen-DAT: I've removed this check. Without power (or == 0) it now enters validateFuelBased(), where it does the check and add the error if needed

source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
```diff
@@ -65,30 +65,43 @@ private boolean validateStandardOffRoad(final StandardOffRoadMobileSource subSou
 
   private boolean validateOffRoadProperties(final StandardOffRoadMobileSource subSource) {
     // Combine all validations in separate statements to make sure each validation is run to collect all validation warnings/errors.
-    boolean valid = validatePowerOrLiterFuel(subSource);
-    valid = validateOffRoadPowerRange(subSource) && valid;
+    final boolean usesUMethod = subSource.getPower() != null && subSource.getPower() > 0;
```
BertScholten: Think you still need to check if the category even has emission factors for power-based?

If you now create a GML with power for a category that does not expect power to be used, with the current code you end up in `validatePowerBased`, and the `expectsPower` method is never called. Slightly surprised the tests don't cover that case yet, perhaps a good one to add?

Might be an idea to have a first validation that determines which method to use, can perhaps use a class-private enum for that with `POWER`, `FUEL`, or `MISSING` as options.  If it returns `MISSING`, do the `MOBILE_SOURCE_MISSING_POWER_OR_LITER_FUEL` error and consider it invalid, if it returns `POWER` do the power based method, and if it returns `FUEL`... It's a bit explicit that way, but might make it more readable.

BertScholten (changes requested):

source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
MichielJanssen-DAT: I've implemented the enum in a new commit, which causes "MISSING" when there is power, but it is not expected, or the other way around. MISSING then causes `ImaerExceptionReason.MOBILE_SOURCE_MISSING_POWER_OR_LITER_FUEL` to be added to the errors.

source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
```diff
@@ -65,30 +69,57 @@ private boolean validateStandardOffRoad(final StandardOffRoadMobileSource subSou
...
+  }
+
+  private EmissionMethod determineEmissionMethod(final StandardOffRoadMobileSource subSource) {
+    final String code = subSource.getOffRoadMobileSourceCode();
+    final boolean hasPower = subSource.getPower() != null && subSource.getPower() > 0;
+    // A category can accept power or fuel. At least one is mandatory.
+    if (hasPower) {
+      return validationHelper.expectsPower(code) ? EmissionMethod.POWER : EmissionMethod.MISSING;
+    }
+    return validationHelper.expectsPower(code) && !validationHelper.expectsLiterFuelPerYear(code) ? EmissionMethod.MISSING : EmissionMethod.FUEL;
```
BertScholten:
```suggestion
    final boolean hasFuel = subSource.getLiterFuelPerYear() != null && subSource.getLiterFuelPerYear() > 0;
    // A category can accept power or fuel. At least one is mandatory.
    if (hasPower && validationHelper.expectsPower(code)) {
      return EmissionMethod.POWER;
    } else if (hasFuel && validationHelper.expectsLiterFuelPerYear(code)) {
      return EmissionMethod.FUEL;
    } else {
      return EmissionMethod.MISSING;
    }
```

source/imaer-util/src/main/java/nl/overheid/aerius/validation/OffRoadValidator.java
```diff
@@ -118,19 +149,14 @@ private boolean validateOffRoadOperatingHours(final StandardOffRoadMobileSource
...
-    if (validationHelper.expectsLiterAdBluePerYear(subSource.getOffRoadMobileSourceCode())) {
-      if (subSource.getLiterAdBluePerYear() == null) {
-        getErrors().add(new AeriusException(ImaerExceptionReason.MOBILE_SOURCE_MISSING_LITER_ADBLUE, subSource.getDescription()));
-        valid = false;
-      } else if (subSource.getLiterFuelPerYear() != null) {
-        validateAdBlueFuelRatio(subSource);
-      }
-    } else if (subSource.getLiterAdBluePerYear() != null && subSource.getLiterAdBluePerYear() == 0) {
+  private void validateOffRoadLiterAdBlue(final StandardOffRoadMobileSource subSource) {
+    // No need for null check, as validateFuelBased() handles that if required
```
BertScholten: comment doesn't really seem to make sense with it checking null anyway? Tbh I think you could let this method stay as it was and do the old `valid = validateOffRoadLiterAdBlue(subSource) && valid` bit?

BertScholten (changes requested):

source/imaer-util/src/test/java/nl/overheid/aerius/validation/OffRoadValidatorTest.java
```diff
@@ -151,22 +151,35 @@ void testSubSourceMissingPower(final Integer fuel) {
...
+
+    mockCategory(MockCategory.FUEL, MockCategory.HOURS, MockCategory.ADBLUE);
+
+    assertValidate(source, 1, 0);
+    assertFirstError(ImaerExceptionReason.MOBILE_SOURCE_MISSING_POWER_OR_LITER_FUEL, List.of(SUB_SOURCE_DESCRIPTION));
+  }
+
+  @Test
+  void testSubSourcePowerSuppliedButNotExpectedShouldBeInvalid() {
+    final OffRoadMobileEmissionSource source = createSource(200, null, 3_000, 500);
```
BertScholten: Think previous parameterizedtest did the same?

source/imaer-util/src/test/java/nl/overheid/aerius/validation/OffRoadValidatorTest.java
```diff
@@ -190,7 +203,7 @@ void testSubSourceMissingAll() {
     final OffRoadMobileEmissionSource source = createSource(null, null, null, null);
 
     mockCategory(MockCategory.POWER, MockCategory.FUEL, MockCategory.HOURS, MockCategory.ADBLUE);
-    assertValidate(source, 3, 0);
+    assertValidate(source, 2, 0);
```
BertScholten, line 212: Guess it's missing the adblue warning now? Which should be OK, just checking I got it right.

BertScholten (approved):
LGTM

source/imaer-util/src/test/java/nl/overheid/aerius/validation/OffRoadValidatorTest.java
MichielJanssen-DAT: Yeah, I've reverted it in a new commit

source/imaer-util/src/test/java/nl/overheid/aerius/validation/OffRoadValidatorTest.java
MichielJanssen-DAT: Yeah. In a new commit I'm also checking for the 2 specific errors

